PR #308, opened 2020-07-22
cuSolverMg wrappers

kshyatt:
This is pretty preliminary, the `getrf` is having access violations I don't understand yet and the accuracy is pretty bad for 32-bit types. This is sitting on top of https://github.com/JuliaGPU/CUDA.jl/pull/306 because they share some multiGPU functionality.


lib/cudalibmg/libcudalibmg.jl
```diff
@@ -0,0 +1,54 @@
+# Automatically-generated headers for cudalibmg
+#
+# DO NOT EDIT THIS FILE DIRECTLY.
+
+
+# Julia wrapper for header: cudalibmg.h
```
maleadt: wait, is there an actual cudalibmg? I thought this was a copy/paste error?

test/cusolver/cusolvermg.jl
```diff
@@ -0,0 +1,114 @@
+using LinearAlgebra, Test
+using CUDA
+using CUDA.CUSOLVER
+
+import LinearAlgebra: BlasInt
+
+@testset "CUSOLVERMG" begin
```
maleadt: we don't do a global testset anymore, since the test infra adds one itself

test/cusolver/cusolvermg.jl
```diff
@@ -0,0 +1,114 @@
...
+
+m = 256
+n = 512
+devs = collect(devices())
+
+if CUDA.has_cusolvermg()
```
maleadt: similarly, this check should be moved to runtests.jl, where we also test for e.g. cudnn

lib/cudalibmg/libcudalibmg.jl
kshyatt: There is in the CUBLASMG code ... maybe this has been hidden away somewhere in the DLL? Most of this code will only be called if you download and use CUBLASMG

kshyatt:
Really having trouble with this `free` problem.

kshyatt:
Tests passed? WHAT?

kshyatt:
The `multigpu` tests show the error while freeing. Anyone have any thoughts about it?

maleadt:
Maybe we should keep support to cusolverMg; cublasMg is early access, latest update 2019/08, and only supports CUDA 10.1... And it looks like NVIDIA hasn't exactly decided on what to do with cudalibmg.

kshyatt:
we do have to keep some of `cudalibmg` though because some of its types (like the grids) are needed for `cusolverMg`

lib/cusolver/wrappers.jl
```diff
@@ -964,3 +976,249 @@ for (jname, fname, elty) in ((:potrfBatched!, :cusolverDnSpotrfBatched, :Float32
         end
     end
 end
+
+function mg_syevd!(jobz::Char, uplo::Char, A; devs=[0], dev_rows=1, dev_cols=length(devs)) # one host-side array A
```
maleadt: why the distinction between `dev_cols=length(devs)`, and `ndevs`?

lib/cudalibmg/wrappers.jl
```diff
@@ -0,0 +1,101 @@
...
+        mat_cpu_bufs[di] = Array(mat[mat_row_inds, mat_col_inds])
+    end
+    for (di, dev) in enumerate(deviceIdsGrid)
+        device!(dev)
+        if !isassigned(streams, di)
+            streams[di] = CuStream()
+        end
+        mat_gpu_buf = CuMatrix{eltype(mat)}(undef, size(mat))
+        unsafe_copyto!(pointer(mat_gpu_buf), pointer(mat_cpu_bufs[di]), length(mat_cpu_bufs[di]), stream = streams[di], async = true)
+        mat_buffers[di] = convert(CuPtr{Cvoid}, pointer(mat_gpu_buf))
```
maleadt: after this point, the array can be collected, right? that probably explains some of the memory errors

maleadt (commented):
I'm looking into this PR, but there's some things unclear.

lib/cusolver/multigpu.jl
```diff
@@ -0,0 +1,357 @@
...
+    current_dev = device()
+    cpu_bufs = Vector{Matrix{eltype(D)}}(undef, ndevices())
+    for (di, dev) in enumerate(devices())
+        device!(dev)
+        dev_row = mod(di - 1, n_row_devs) + 1
+        dev_col = div(di - 1, n_row_devs) + 1
+        row_inds = ((dev_row-1)*row_block_size+1):min(dev_row*row_block_size, size(D, 1))
+        col_inds = ((dev_col-1)*col_block_size+1):min(dev_col*col_block_size, size(D, 2))
+        cpu_bufs[di] = Matrix{eltype(D)}(undef, length(row_inds), length(col_inds))
+        unsafe_copyto!(cpu_bufs[di], 1, dDs[di], 1, length(cpu_bufs[di]))
```
maleadt: if I change this to an ordinary copyto, stuff breaks; is it expected that the length of dDs[di] doesn't match length(cpu_bufs[di])?

maleadt:
Passes tests :tada:

kshyatt:
Thanks for rescuing this!

maleadt:
Any thoughts on those inline comments I had?

maleadt:
Bump.

lib/cusolver/multigpu.jl
kshyatt: I think so yes

kshyatt: Just checked to confirm, yeah the tests fail with `copyto!`

kshyatt: Update: fixed this.

kshyatt:
Finally done!!!

PR #321, opened 2020-07-26
Performance improvements by calling cuDNN API

### LINKED ISSUE #318: Performance: cudnn algorithm selection
There is a number of algorithm selection utilities in CUDNN to choose the most performant algorithm for a given input size for convolution, rnn etc. These should be integrated at least to the NNlib level interface to get the best performance. Some of these are heuristic (the *Get* functions) and do not work very well (I think they were deprecated in cudnn 8.0). Some of them actually try the various algorithms on the given input sizes and actually help performance significantly (the *Find* functions). There are two issues with the versions that try algorithms: (1) results should be cached so we do not try again for the same size input, this needs to be managed with multiple GPUs, multi-threading etc, (2) different algorithms require different amounts of workspace, we need to control the allocation of this workspace and balance the extra memory cost with the potential speed benefit. Here is a list of current algorithm selection utilities:
```
cudnnFindConvolutionForwardAlgorithm
cudnnFindConvolutionForwardAlgorithmEx
cudnnGetConvolutionForwardAlgorithm
cudnnGetConvolutionForwardAlgorithm_v7
cudnnFindConvolutionBackwardFilterAlgorithm
cudnnFindConvolutionBackwardFilterAlgorithmEx
cudnnGetConvolutionBackwardFilterAlgorithm
cudnnGetConvolutionBackwardFilterAlgorithm_v7
cudnnFindConvolutionBackwardDataAlgorithm
cudnnFindConvolutionBackwardDataAlgorithmEx
cudnnGetConvolutionBackwardDataAlgorithm
cudnnGetConvolutionBackwardDataAlgorithm_v7
cudnnFindRNNForwardInferenceAlgorithmEx
cudnnFindRNNForwardTrainingAlgorithmEx
cudnnFindRNNBackwardDataAlgorithmEx
cudnnFindRNNBackwardWeightsAlgorithmEx
```

gartangh:
* Selects the fastest convolutional algorithm
* Faster activations, if supported
* Faster softmax
* Merges convolution, bias addition, and activation function if possible

Depends on:
https://github.com/FluxML/NNlib.jl/pull/228

### DIFF
--- a/lib/cudnn/conv.jl
+++ b/lib/cudnn/conv.jl
@@ -38,6 +38,10 @@ function ConvDesc(T, N, padding, stride, dilation, mode)
     else
         cudnnSetConvolutionNdDescriptor(cd[],N,cdsize(padding,N),cdsize(stride,N),cdsize(dilation,N),mode)
     end
+    # enable tensor math mode if our device supports it
+    if capability(device()) >= v"7.0" && version() >= v"9"
+        cudnnSetConvolutionMathType(cd[], CUDNN_TENSOR_OP_MATH)
+    end
     this = ConvDesc(cd[])
     finalizer(unsafe_free!, this)
     return this
@@ -54,6 +58,90 @@ end
 
 
 # wrappers
+
+# Forward
+
+function cudnnGetConvolutionForwardAlgorithmMaxCount()
+    count=@argout(
+        cudnnGetConvolutionForwardAlgorithmMaxCount(
+            handle(),
+            out(Ref{Cint}()))
+        )[]
+    return count
+end
+
+# will be removed in cuDNN 8
+function cudnnGetConvolutionForwardAlgorithm(y::CuArray{T,N}, x::CuArray{T,N}, w::CuArray{T,N},
+                                             cdims::DenseConvDims; preference=0, workspacesize=1<<32) where {T,N}
+    algo=@argout(
+        cudnnGetConvolutionForwardAlgorithm(
+            handle(), TensorDesc(x),
+            FilterDesc(w), ConvDesc(T, cdims),
+            TensorDesc(y),
+            cudnnConvolutionFwdPreference_t(preference),
+            Csize_t(workspacesize),
+            out(Ref{cudnnConvolutionFwdAlgo_t}()))
+        )[]
+    return algo
+end
+
+function cudnnGetConvolutionForwardAlgorithm_v7(y::CuArray{T,N}, x::CuArray{T,N}, w::CuArray{T,N},
+                                                cdims::DenseConvDims; count=-1) where {T,N}
+    if count < 0
+        count = cudnnGetConvolutionForwardAlgorithmMaxCount()
+    end
+    perfResults = Array{cudnnConvolutionFwdAlgoPerf_t, 1}(undef, count)
+    returnedAlgoCount=@argout(
+        cudnnGetConvolutionForwardAlgorithm_v7(
+            handle(), TensorDesc(x),
+            FilterDesc(w), ConvDesc(T, cdims),
+            TensorDesc(y),
+            Cint(count),
+            out(Ref{Cint}()),
+            perfResults)
+        )[]
+    return returnedAlgoCount, perfResults
+end
+
+function cudnnFindConvolutionForwardAlgorithm(y::CuArray{T,N}, x::CuArray{T,N}, w::CuArray{T,N},
+                                              cdims::DenseConvDims; count=-1) where {T,N}
+    if count < 0
+        count = cudnnGetConvolutionForwardAlgorithmMaxCount()
+    end
+    perfResults = Array{cudnnConvolutionFwdAlgoPerf_t, 1}(undef, count)
+    returnedAlgoCount=@argout(
+        cudnnFindConvolutionForwardAlgorithm(
+            handle(), TensorDesc(x),
+            FilterDesc(w), ConvDesc(T, cdims),
+            TensorDesc(y),
+            Cint(count),
+            out(Ref{Cint}()),
+            perfResults)
+        )[]
+    return returnedAlgoCount, perfResults
+end
+
+function cudnnFindConvolutionForwardAlgorithmEx(y::CuArray{T,N}, x::CuArray{T,N}, w::CuArray{T,N},
+                                              cdims::DenseConvDims; count=-1, workspacesize=1<<32) where {T,N}
+    if count < 0
+        count = cudnnGetConvolutionForwardAlgorithmMaxCount()
+    end
+    @workspace size=workspacesize workspace->begin
+        perfResults = Array{cudnnConvolutionFwdAlgoPerf_t, 1}(undef, count)
+        returnedAlgoCount=@argout(
+            cudnnFindConvolutionForwardAlgorithmEx(
+                handle(), TensorDesc(x), x,
+                FilterDesc(w), w, ConvDesc(T, cdims),
+                TensorDesc(y), y,
+                Cint(count),
+                out(Ref{Cint}()),
+                perfResults,
+                workspace,
+                workspacesize)
+            )[]
+        return returnedAlgoCount, perfResults
+    end
+end
 
 function cudnnConvolutionForward(y::CuArray{T,N}, x::CuArray{T,N}, w::CuArray{T,N},
                                  cdims::DenseConvDims; algo=0, alpha=1, beta=0) where {T,N}
@@ -73,6 +161,100 @@ function cudnnConvolutionForward(y::CuArray{T,N}, x::CuArray{T,N}, w::CuArray{T,
     return y
 end
 
+function cudnnConvolutionBiasActivationForward(y::CuArray{T,N}, x::CuArray{T,N}, w::CuArray{T,N}, z::CuArray{T,N}, bias::CuArray{T,N},
+                                               cdims::DenseConvDims; algo=0, alpha1=1, alpha2=1,
+                                               activationMode=CUDNN_ACTIVATION_RELU, activationCoeff=0.0, activationReluNanOpt=CUDNN_NOT_PROPAGATE_NAN) where {T,N}
+    @workspace size=@argout(
+            cudnnGetConvolutionForwardWorkspaceSize(
+                handle(), TensorDesc(x),
+                FilterDesc(w), ConvDesc(T, cdims),
+                TensorDesc(y),
+                cudnnConvolutionFwdAlgo_t(algo),
+                out(Ref{Csize_t}()))
+        )[] workspace->begin
+            cudnnConvolutionBiasActivationForward(
+                handle(), Ref(T(alpha1)), TensorDesc(x), x, FilterDesc(w), w,
+                ConvDesc(T, cdims), cudnnConvolutionFwdAlgo_t(algo), workspace,
+                sizeof(workspace), Ref(T(alpha2)), TensorDesc(z), z, TensorDesc(bias), bias, ActivationDesc(activationMode, activationCoeff, activationReluNanOpt), TensorDesc(y),y)
+        end
+    return y
+end
+
+# Backward data
+
+function cudnnGetConvolutionBackwardDataAlgorithmMaxCount()
+    count=@argout(
+        cudnnGetConvolutionBackwardDataAlgorithmMaxCount(
+            handle(),
+            out(Ref{Cint}()))
+        )[]
+    return count
+end
+
+# will be removed in cuDNN 8
+function cudnnGetConvolutionBackwardDataAlgorithm(dx::CuArray{T,N}, w::CuArray{T,N}, dy::CuArray{T,N},
+                                                  cdims::DenseConvDims; preference=0, workspacesize=1<<32) where {T,N}
+    algo=@argout(
+        cudnnGetConvolutionBackwardDataAlgorithm(
+            handle(), FilterDesc(w), TensorDesc(dy), ConvDesc(T, cdims),
+            TensorDesc(dx), cudnnConvolutionBwdDataPreference_t(preference),
+            Csize_t(workspacesize), out(Ref{cudnnConvolutionBwdDataAlgo_t}()))
+        )[]
+    return algo
+end
+
+function cudnnGetConvolutionBackwardDataAlgorithm_v7(dx::CuArray{T,N}, w::CuArray{T,N}, dy::CuArray{T,N},
+                                                     cdims::DenseConvDims; count=-1) where {T,N}
+    if count < 0
+        count = cudnnGetConvolutionBackwardDataAlgorithmMaxCount()
+    end
+    perfResults = Array{cudnnConvolutionBwdDataAlgoPerf_t, 1}(undef, count)
+    returnedAlgoCount=@argout(
+        cudnnGetConvolutionBackwardDataAlgorithm_v7(
+            handle(), FilterDesc(w), TensorDesc(dy),
+            ConvDesc(T, cdims), TensorDesc(dx),
+            Cint(count),
+            out(Ref{Cint}()), perfResults)
+        )[]
+    return returnedAlgoCount, perfResults
+end
+
+function cudnnFindConvolutionBackwardDataAlgorithm(dx::CuArray{T,N}, w::CuArray{T,N}, dy::CuArray{T,N},
+                                                   cdims::DenseConvDims; count=-1) where {T,N}
+    if count < 0
+        count = cudnnGetConvolutionBackwardDataAlgorithmMaxCount()
+    end
+    perfResults = Array{cudnnConvolutionBwdDataAlgoPerf_t, 1}(undef, count)
+    returnedAlgoCount=@argout(
+        cudnnFindConvolutionBackwardDataAlgorithm(
+            handle(), FilterDesc(w), TensorDesc(dy),
+            ConvDesc(T, cdims), TensorDesc(dx),
+            Cint(count),
+            out(Ref{Cint}()), perfResults)
+        )[]
+    return returnedAlgoCount, perfResults
+end
+
+function cudnnFindConvolutionBackwardDataAlgorithmEx(dx::CuArray{T,N}, w::CuArray{T,N}, dy::CuArray{T,N},
+                                                     cdims::DenseConvDims; count=-1, workspacesize=1<<32) where {T,N}
+    if count < 0
+        count = cudnnGetConvolutionBackwardDataAlgorithmMaxCount()
+    end
+    @workspace size=workspacesize workspace->begin
+        perfResults = Array{cudnnConvolutionBwdDataAlgoPerf_t, 1}(undef, count)
+        returnedAlgoCount=@argout(
+            cudnnFindConvolutionBackwardDataAlgorithmEx(
+                handle(), FilterDesc(w), w, TensorDesc(dy), dy,
+                ConvDesc(T, cdims), TensorDesc(dx), dx,
+                Cint(count),
+                out(Ref{Cint}()),
+                perfResults, workspace,
+                workspacesize)
+            )[]
+        return returnedAlgoCount, perfResults
+    end
+end
+
 function cudnnConvolutionBackwardData(dx::CuArray{T,N}, w::CuArray{T,N}, dy::CuArray{T,N},
                                       cdims::DenseConvDims; algo=0, alpha=1, beta=0) where {T,N}
     @workspace size=@argout(
@@ -90,6 +272,83 @@ function cudnnConvolutionBackwardData(dx::CuArray{T,N}, w::CuArray{T,N}, dy::CuA
                 Ref{T}(beta), TensorDesc(dx), dx)
         end
     return dx
+end
+
+# Backward filter
+
+function cudnnGetConvolutionBackwardFilterAlgorithmMaxCount()
+    count=@argout(
+        cudnnGetConvolutionBackwardFilterAlgorithmMaxCount(
+            handle(),
+            out(Ref{Cint}()))
+        )[]
+    return count
+end
+
+# will be removed in cuDNN 8
+function cudnnGetConvolutionBackwardFilterAlgorithm(dw::CuArray{T,N}, x::CuArray{T,N}, dy::CuArray{T,N},
+                                                    cdims::DenseConvDims; preference=0, workspacesize=1<<32) where {T,N}
+    algo=@argout(
+        cudnnGetConvolutionBackwardFilterAlgorithm(
+            handle(), TensorDesc(x), TensorDesc(dy),
+            ConvDesc(T, cdims), FilterDesc(dw), cudnnConvolutionBwdFilterPreference_t(preference),
+            Csize_t(workspacesize), out(Ref{cudnnConvolutionBwdFilterAlgo_t}()))
+        )[]
+    return algo
+end
+
+function cudnnGetConvolutionBackwardFilterAlgorithm_v7(dw::CuArray{T,N}, x::CuArray{T,N}, dy::CuArray{T,N},
+                                                     cdims::DenseConvDims; count=-1) where {T,N}
+    if count < 0
+        count = cudnnGetConvolutionBackwardFilterAlgorithmMaxCount()
+    end
+    perfResults = Array{cudnnConvolutionBwdFilterAlgoPerf_t, 1}(undef, count)
+    returnedAlgoCount=@argout(
+        cudnnGetConvolutionBackwardFilterAlgorithm_v7(
+            handle(), TensorDesc(x), TensorDesc(dy),
+            ConvDesc(T, cdims), FilterDesc(dw),
+            Cint(count),
+            out(Ref{Cint}()),
+            perfResults)
+        )[]
+    return returnedAlgoCount, perfResults
+end
+
+function cudnnFindConvolutionBackwardFilterAlgorithm(dw::CuArray{T,N}, x::CuArray{T,N}, dy::CuArray{T,N},
+                                                     cdims::DenseConvDims; count=-1) where {T,N}
+    if count < 0
+        count = cudnnGetConvolutionBackwardFilterAlgorithmMaxCount()
+    end
+    perfResults = Array{cudnnConvolutionBwdFilterAlgoPerf_t, 1}(undef, count)
+    returnedAlgoCount=@argout(
+        cudnnFindConvolutionBackwardFilterAlgorithm(
+            handle(), TensorDesc(x), TensorDesc(dy),
+            ConvDesc(T, cdims), FilterDesc(dw),
+            Cint(count),
+            out(Ref{Cint}()),
+            perfResults)
+        )[]
+    return returnedAlgoCount, perfResults
+end
+
+function cudnnFindConvolutionBackwardFilterAlgorithmEx(dw::CuArray{T,N}, x::CuArray{T,N}, dy::CuArray{T,N},
+                                                     cdims::DenseConvDims; count=-1, workspacesize=1<<32) where {T,N}
+    if count < 0
+        count = cudnnGetConvolutionBackwardFilterAlgorithmMaxCount()
+    end
+    @workspace size=workspacesize workspace->begin
+        perfResults = Array{cudnnConvolutionBwdFilterAlgoPerf_t, 1}(undef, count)
+        returnedAlgoCount=@argout(
+            cudnnFindConvolutionBackwardFilterAlgorithmEx(
+                handle(), TensorDesc(x), x, TensorDesc(dy),
+                dy, ConvDesc(T, cdims), FilterDesc(dw), dw,
+                Cint(count),
+                out(Ref{Cint}()),
+                perfResults, workspace,
+                workspacesize)
+            )[]
+        return returnedAlgoCount, perfResults
+    end
 end
 
 function cudnnConvolutionBackwardFilter(dw::CuArray{T,N}, x::CuArray{T,N}, dy::CuArray{T,N},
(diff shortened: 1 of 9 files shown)


lib/cudnn/conv.jl
```diff
@@ -38,6 +38,9 @@ function ConvDesc(T, N, padding, stride, dilation, mode)
     else
         cudnnSetConvolutionNdDescriptor(cd[],N,cdsize(padding,N),cdsize(stride,N),cdsize(dilation,N),mode)
     end
+    if version() >= v"7"
+        cudnnSetConvolutionMathType(cd[], cudnnMathType_t(1))
```
maleadt: That's, CUDNN_TENSOR_OP_MATH, right? We can't use that unconditionally since it doesn't have the same accuracy. Instead: https://github.com/JuliaGPU/CUDA.jl/blob/463a41295bfede5125c584e6be9c51a4b9074e12/lib/cublas/CUBLAS.jl#L42-L46

maleadt:
cc @denizyuret 
Also see https://github.com/JuliaGPU/CUDA.jl/issues/318, looks like the Find* methods would be even better, but this is a good start.

lib/cudnn/conv.jl
gartangh: Yes, that is correct. I have added the extra conditions in this commit: https://github.com/JuliaGPU/CUDA.jl/pull/321/commits/480ea6fdb2edcca46152cd0398d7b6e456b748d7.
I have also added a condition for using the fast version of softmax over the accurate one.

denizyuret:
1. Can we also do algo selection for RNNs? 
2. And an we make the fast modes always the default?
3. To use the Find* methods I have a bunch of design issues listed in #318, I am not happy with my Knet implementation. If you have suggestions, I'd be happy to implement.

gartangh:
@maleadt @denizyuret 
I just implemented wrappers for:

* `cudnnGetConvolutionForwardAlgorithm_v7`
* `cudnnFindConvolutionForwardAlgorithm`
* `cudnnFindConvolutionForwardAlgorithmEx`
* `cudnnGetConvolutionForwardAlgorithmMaxCount`

1.  in cuDNN 8, the following functions will be deprecated:

    * `cudnnFindRNNForwardInferenceAlgorithmEx`
    * `cudnnFindRNNForwardTrainingAlgorithmEx`
    * `cudnnFindRNNBackwardDataAlgorithmEx`
    * `cudnnFindRNNBackwardWeightsAlgorithmEx`

    so I do not think it is worth the effort to implement them at the moment.
2. I am actually in favor of using accurate algorithms over fast ones by default.
3. Any suggestions on how to cache the algorithm for certain input sizes? And how to limit memory usage? I am currently proposing `cudnnGetConvolutionForwardAlgorithm_v7` until we have a way for caching, as `cudnnGetConvolutionForwardAlgorithm` will be removed in cuDNN 8. This can however still take algorithms that use lots of memory.

denizyuret:
> @maleadt @denizyuret
> I just implemented wrappers for:
> 
> * `cudnnGetConvolutionForwardAlgorithm_v7`
> * `cudnnFindConvolutionForwardAlgorithm`
> * `cudnnFindConvolutionForwardAlgorithmEx`
> * `cudnnGetConvolutionForwardAlgorithmMaxCount`

Great! 

> 
> 1. in cuDNN 8, the following functions will be deprecated:
>    
>    * `cudnnFindRNNForwardInferenceAlgorithmEx`
>    * `cudnnFindRNNForwardTrainingAlgorithmEx`
>    * `cudnnFindRNNBackwardDataAlgorithmEx`
>    * `cudnnFindRNNBackwardWeightsAlgorithmEx`
>    
>    so I do not think it is worth the effort to implement them at the moment.

Good catch, I agree.

> 2. I am actually in favor of using accurate algorithms over fast ones by default.

What is your reasoning here? Making accurate algorithms the default may keep us behind in benchmarks and discourage people for adapting Julia as their deep learning language of choice. Accuracy is almost never important in deep learning (I have worked with 8-bit floats with little degradation in performance) and low accuracy sometimes helps as a regularizer and improves generalization.

> 3. Any suggestions on how to cache the algorithm for certain input sizes? And how to limit memory usage? I am currently proposing `cudnnGetConvolutionForwardAlgorithm_v7` until we have a way for caching, as `cudnnGetConvolutionForwardAlgorithm` will be removed in cuDNN 8. This can however still take algorithms that use lots of memory.

As mentioned in https://github.com/JuliaGPU/CUDA.jl/issues/318 my experience with Knet is:
* "Get" functions did not have very good heuristics and I could not beat the benchmarks with them. Things may have improved since then, this was a couple years ago.
* "Find" functions are better, they try each possibility after all. However as you mentioned they require caching and workspace memory.
* `cudnnFindConvolutionForwardAlgorithm()` does the required memory allocation for its experiments on its own which makes things simpler. But if the CUDA pool does not leave enough memory for an external cudaMalloc you risk running out of memory and getting CUDNN_STATUS_ALLOC_FAILED.
* Thus my current implementation uses `cudnnFindConvolutionForwardAlgorithmEx()` which lets you allocate the workspace memory (which when you do using CuArrays avoid the out of memory problem). However this makes things a bit complicated because you need to figure out how much workspace memory to allocate for a given array size to try all the algorithms you want.
* I cache results in global dictionaries (input sizes -> algo). This is probably not good for multi-threading, multi-gpu etc. and needs to be fixed. I have a maximum number of keys I cache and just return the default algo (0) if more requests keep coming.

The code is in https://github.com/denizyuret/Knet.jl/blob/master/src/conv.jl

gartangh:
@denizyuret 
I also added wrappers for these functions, as they are fairly similar to the once I implemented before:
* `cudnnFindConvolutionBackwardFilterAlgorithm`
* `cudnnFindConvolutionBackwardFilterAlgorithmEx`
* `cudnnGetConvolutionBackwardFilterAlgorithm`
* `cudnnGetConvolutionBackwardFilterAlgorithm_v7`
* `cudnnGetConvolutionBackwardFilterAlgorithmMaxCount`
* `cudnnFindConvolutionBackwardDataAlgorithm`
* `cudnnFindConvolutionBackwardDataAlgorithmEx`
* `cudnnGetConvolutionBackwardDataAlgorithm`
* `cudnnGetConvolutionBackwardDataAlgorithm_v7`
* `cudnnGetConvolutionBackwardDataAlgorithmMaxCount`

gartangh:
@denizyuret 
> What is your reasoning here? Making accurate algorithms the default may keep us behind in benchmarks and discourage people for adapting Julia as their deep learning language of choice. Accuracy is almost never important in deep learning (I have worked with 8-bit floats with little degradation in performance) and low accuracy sometimes helps as a regularizer and improves generalization.

I feel like Julia and its packages always favor accuracy over performance. So for the sake of consistency, we could do the same here. I completely agree with your reasoning that accuracy is not that important for Deep Learning, so I would be happy to use the most performant code by default, but that is not a decision I want to make by myself.

denizyuret:
> @denizyuret
> 
> > What is your reasoning here? Making accurate algorithms the default may keep us behind in benchmarks and discourage people for adapting Julia as their deep learning language of choice. Accuracy is almost never important in deep learning (I have worked with 8-bit floats with little degradation in performance) and low accuracy sometimes helps as a regularizer and improves generalization.
> 
> I feel like Julia and its packages always favor accuracy over performance. So for the sake of consistency, we could do the same here. I completely agree with your reasoning that accuracy is not that important for Deep Learning, so I would be happy to use the most performant code by default, but that is not a decision I want to make by myself.

@ViralBShah what do you think? Can we break with tradition and make the most performant variants of CUDNN algorithms the default for the sake of deep learning packages? We can add keyword options to choose the more precise algorithms and document them, but I think it is important (esp. for new users) that the defaults work fast.

gartangh:
@maleadt @denizyuret 
>As mentioned in #318 my experience with Knet is:
"Get" functions did not have very good heuristics and I could not beat the benchmarks with them. Things may have improved since then, this was a couple years ago.
"Find" functions are better, they try each possibility after all. However as you mentioned they require caching and workspace memory.
cudnnFindConvolutionForwardAlgorithm() does the required memory allocation for its experiments on its own which makes things simpler. But if the CUDA pool does not leave enough memory for an external cudaMalloc you risk running out of memory and getting CUDNN_STATUS_ALLOC_FAILED.
Thus my current implementation uses cudnnFindConvolutionForwardAlgorithmEx() which lets you allocate the workspace memory (which when you do using CuArrays avoid the out of memory problem). However this makes things a bit complicated because you need to figure out how much workspace memory to allocate for a given array size to try all the algorithms you want.
I cache results in global dictionaries (input sizes -> algo). This is probably not good for multi-threading, multi-gpu etc. and needs to be fixed. I have a maximum number of keys I cache and just return the default algo (0) if more requests keep coming.
The code is in https://github.com/denizyuret/Knet.jl/blob/master/src/conv.jl

I just added a way of caching the algorithm returned from `cudnnFindConvolutionForwardAlgorithmEx`, similar to how you did it in Knet.jl. I opted for a dictionary with limited capacity and Least Recently Used (LRU) replacement. I would love to have some feedback on this. (Tomorrow, I will run some benchmarks and profile a Deep Neural Network to measure the differences.)

denizyuret:
> @maleadt @denizyuret
> 
> > As mentioned in #318 my experience with Knet is:
> > "Get" functions did not have very good heuristics and I could not beat the benchmarks with them. Things may have improved since then, this was a couple years ago.
> > "Find" functions are better, they try each possibility after all. However as you mentioned they require caching and workspace memory.
> > cudnnFindConvolutionForwardAlgorithm() does the required memory allocation for its experiments on its own which makes things simpler. But if the CUDA pool does not leave enough memory for an external cudaMalloc you risk running out of memory and getting CUDNN_STATUS_ALLOC_FAILED.
> > Thus my current implementation uses cudnnFindConvolutionForwardAlgorithmEx() which lets you allocate the workspace memory (which when you do using CuArrays avoid the out of memory problem). However this makes things a bit complicated because you need to figure out how much workspace memory to allocate for a given array size to try all the algorithms you want.
> > I cache results in global dictionaries (input sizes -> algo). This is probably not good for multi-threading, multi-gpu etc. and needs to be fixed. I have a maximum number of keys I cache and just return the default algo (0) if more requests keep coming.
> > The code is in https://github.com/denizyuret/Knet.jl/blob/master/src/conv.jl
> 
> I just added a way of caching the algorithm returned from `cudnnFindConvolutionForwardAlgorithmEx`, similar to how you did it in Knet.jl. I opted for a dictionary with limited capacity and Least Recently Used (LRU) replacement. I would love to have some feedback on this. (Tomorrow, I will run some benchmarks and profile a Deep Neural Network to measure the differences.)

That's great. About the LRU: my purpose in limiting the number was not to conserve memory: the entries do not take that much space. It was more about not wasting too much time with *Find* calls if the array sizes keep changing, which LRU will not limit. Although in practice this probably never happens so it doesn't matter that much as long as we have a cache size that can handle training common deep models. And I can see the logic of LRU if the user switches between different models during the same session.

lib/cudnn/nnlib.jl
```diff
@@ -67,6 +94,7 @@ fix1d(cdims::DenseConvDims{1,K,C_in,C_out,S,P,D,F}) where {K,C_in,C_out,S,P,D,F}
 fix1d(pdims::PoolDims{1,K,S,P,D}) where {K,S,P,D,F} =
   PoolDims{2,(K...,1),(S...,1),(P...,0,0),(D...,1)}((pdims.I..., 1), pdims.C_in)
 
+conv_forward = CircularDict{Tuple, Int32}(100)
```
maleadt: Why only cache a limited amount? If you key with not much uniqueness, only `(T, size, DenseConvDims)`, we should have a high enough hit rate? The entries are small, too, and don't keep anything alive.

maleadt: Should be `const`

lib/cudnn/nnlib.jl
```diff
@@ -76,11 +104,17 @@ function conv_bias_act!(y::CuArray{T}, x::CuArray{T}, w::CuArray{T}, cdims::Dens
...
   if σ == NNlib.relu
     if algo < 0
-      # algo = UInt32(cudnnGetConvolutionForwardAlgorithm(fix1d(y), fix1d(x), fix1d(w), fix1d(cdims))) # will be removed in cuDNN 8
-      returnedAlgoCount, perfResults = cudnnGetConvolutionForwardAlgorithm_v7(fix1d(y), fix1d(x), fix1d(w), fix1d(cdims))
-      # returnedAlgoCount, perfResults = cudnnFindConvolutionForwardAlgorithm(fix1d(y), fix1d(x), fix1d(w), fix1d(cdims))
-      # returnedAlgoCount, perfResults = cudnnFindConvolutionForwardAlgorithmEx(fix1d(y), fix1d(x), fix1d(w), fix1d(cdims))
-      algo = UInt32(perfResults[1].algo)
+      global conv_forward
+      key = (T, size(x)[end], cdims)
+      algo = get(conv_forward, key, -1)
```
maleadt: `get!` would be nice

lib/cudnn/nnlib.jl
```diff
@@ -76,11 +104,17 @@ function conv_bias_act!(y::CuArray{T}, x::CuArray{T}, w::CuArray{T}, cdims::Dens
...
+      if algo < 0 # not in conv_forward
+        # algo = UInt32(cudnnGetConvolutionForwardAlgorithm(fix1d(y), fix1d(x), fix1d(w), fix1d(cdims))) # will be removed in cuDNN 8
+        # returnedAlgoCount, perfResults = cudnnGetConvolutionForwardAlgorithm_v7(fix1d(y), fix1d(x), fix1d(w), fix1d(cdims))
+        # returnedAlgoCount, perfResults = cudnnFindConvolutionForwardAlgorithm(fix1d(y), fix1d(x), fix1d(w), fix1d(cdims))
+        returnedAlgoCount, perfResults = cudnnFindConvolutionForwardAlgorithmEx(fix1d(y), fix1d(x), fix1d(w), fix1d(cdims), workspacesize=workspacesize(fix1d(x)))
+        algo = UInt32(perfResults[1].algo)
+        push!(conv_forward, key, algo)
```
maleadt: Do other properties of y/x/w not influence the selected algorithm?

gartangh:
@denizyuret 
>That's great. About the LRU: my purpose in limiting the number was not to conserve memory: the entries do not take that much space. It was more about not wasting too much time with Find calls if the array sizes keep changing, which LRU will not limit. Although in practice this probably never happens so it doesn't matter that much as long as we have a cache size that can handle training common deep models. And I can see the logic of LRU if the user switches between different models during the same session.

@maleadt 
>Why only cache a limited amount? If you key with not much uniqueness, only (T, size, DenseConvDims), we should have a high enough hit rate? The entries are small, too, and don't keep anything alive.

I was not really aiming to conserve memory, since (like you say) the entries do not take that much space.
I really was thinking about "what if we switch between large models?".
I think the conclusion from your comments is that we can use a regular dictionary without a capacity limit.

lib/cudnn/nnlib.jl
gartangh: fixed in https://github.com/JuliaGPU/CUDA.jl/pull/321/commits/b7d678727d87e75faa56dbe750288b5b1522c35e for all 3 dictionaries

lib/cudnn/nnlib.jl
gartangh: in https://github.com/JuliaGPU/CUDA.jl/commit/b7d678727d87e75faa56dbe750288b5b1522c35e, I am no longer using get/push methods, but indexing with [key] instead.

lib/cudnn/nnlib.jl
gartangh: I have added the strides of x/y/w in https://github.com/JuliaGPU/CUDA.jl/pull/321/commits/41c7fe5823917477011d7fe253e6fccd66e3b2d2.
The spatial size of x/y/w, the padding, stride, dilation, ... are all inside the `cdims` or can be calculated using only depending on the elements in cdims.
The data type and batch size are not in cdims, so I added those already.

lib/cudnn/nnlib.jl
gartangh: Cf. my comment here: https://github.com/JuliaGPU/CUDA.jl/pull/321#issuecomment-665542278

gartangh:
So I just did the measurements I promised: (all measurements are in milliseconds)

| Network  | Input size | #master | #conv_bias_act | #experimentation |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| VGG19  | (224,224,3,64)  | 263 | 160 | 154 |
| Darknet-19  | (224,224,3,128) | 120 | 90 | 84 |
| ResNet-50  | (224,224,3,64) | 253 | 285 | 197 |
| Darknet-53  | (256,256,3,64) | 194 | 164 | 148 |

The reason that Resnet-50 is slower, is that there are lots of memory issues. I believe that if we can come up with a better heuristic for the `workspacesize`, we can speed up all these networks.

denizyuret:
> The reason that Resnet-50 is slower, is that there are lots of memory issues. I believe that if we can come up with a better heuristic for the `workspacesize`, we can speed up all these networks.

These look very good. Can you give a bit more detail about Resnet-50? Is it the workspacesize used during algorithm discovery, or during training/inference you think is the problem? Is it running close to the GPU memory limit? About workspacesize heuristics, I use two:

(1) Limit the maximum workspacesize during discovery to a reasonable percentage of available memory and/or a reasonable multiple of the input array size: https://github.com/denizyuret/Knet.jl/blob/12430600ea0782cb16cd1aa5887796a5f0d4359f/src/conv.jl#L538

(2) When you are picking an algorithm looking at the *Find* results, prefer ones with lower memory up to a 10% speed penalty: https://github.com/denizyuret/Knet.jl/blob/12430600ea0782cb16cd1aa5887796a5f0d4359f/src/conv.jl#L617

gartangh:
@denizyuret , I was already using your first heuristic but overlooked the second.
I also found some speedups for my implementations of activation functions on the GPU.
ResNet-50 now uses 25 % less memory and clocks in below 200ms.

I have updated my previous comment with new measurements from the experimentation branch.

denizyuret:
@gartangh, please be careful when adding bias, activation functions etc: I have observed inferior performance relative to handwritten kernels in previous versions of cudnn, so I would not make these default without benchmarking.

gartangh:
@denizyuret , as per your request:

Activations

Activations Speedup

Where before is the master branch and after the conv_bias_act branch.
These cuDNN kernels seem to outperform Julia Broadcast kernels by quite some margin.

maleadt:
>  These cuDNN kernels seem to outperform Julia Broadcast kernels by quite some margin.

That's with all the recent broadcast optimizations, I take it?

gartangh:
@maleadt 

>That's with all the recent broadcast optimizations, I take it?

I measured this with this commit on the master branch: https://github.com/JuliaGPU/CUDA.jl/commit/8c61a0556a5bae6096958302cadc6bd3435666b9.
Were there any significant broadcast optimizations in the last 2 days?

maleadt:
>  Were there any significant broadcast optimizations in the last 2 days?

No. Its interesting CUDNN's activation functions still perform so much better. Does the identity operation spawn a kernel that indexes elements and copies them? If so, our indexing is OK, but the actual operation is slow. If not, maybe they amortize indexing across threads (ours is know to be fairly expensive due to the `div`s that arise from the CartesianIndices).

gartangh:
@maleadt 

>Does the identity operation spawn a kernel that indexes elements and copies them?

I was delighted to see that the identity operation does not spawn an unnecessary kernel. This is also the reason why I included it in the bar chart with zeros.

gartangh:
@maleadt

> > Does the identity operation spawn a kernel that indexes elements and copies them?
>
>I was delighted to see that the identity operation does not spawn an unnecessary kernel. This is also the reason why I included it in the bar chart with zeros.

I am going to completely take back my words. Calling `identity.(x)` on CuArrays does spawn an unnecessary Julia_Broadcast kernel. I have already made changes to this branch, to mitigate that problem. Now, no more Julia_Broadcast kernels are called when running Darknet-19, for example.

gartangh:
@maleadt @denizyuret 

I just had the best benchmark so far for ResNet-50: 189 ms (253 on #master)! However, Darknet-19 is suddenly super slow: 160 ms (120 on #master).
When I profile with Nsight Systems, the problem becomes very clear: memory management.
I believe that without this problem, we can achieve sub 100 ms for ResNet-50.

Below, I added a screenshot from a part of ResNet-50.
The kernels in blue only take up a small portion of the time, there are very large gaps between them where the GPU sits idle.
I am not running out of memory either, only 10 GB / 16 GB was allocated during the complete execution of this network.
Do you have any ideas on how to solve this? I cannot seem to figure out where all these allocations come from.

Screenshot from 2020-08-02 00-32-41

denizyuret:
Something is not quite right with `cudnnAddTensor`:
```
julia> a = CUDA.rand(2,3,4,5);                                                                                                                                               
                                                                                                                                                                             
julia> b = CUDA.rand(2,1,1,5);                                                                                                                                               
                                                                                                                                                                             
julia> a .+ b;                                                                                                                                                               
ERROR: CUDNNError: CUDNN_STATUS_NOT_SUPPORTED (code 9)                                                                                                                       
Stacktrace:                                                                                                                                                                  
 [1] throw_api_error(::CUDA.CUDNN.cudnnStatus_t) at /home/deniz/.julia/dev/CUDA/lib/cudnn/error.jl:19                                                                        
 [2] macro expansion at /home/deniz/.julia/dev/CUDA/lib/cudnn/error.jl:30 [inlined]                                                                                          
 [3] cudnnAddTensor(::Ptr{Nothing}, ::Base.RefValue{Float32}, ::CUDA.CUDNN.TensorDesc, ::CuArray{Float32,4}, ::Base.RefValue{Float32}, ::CUDA.CUDNN.TensorDesc, ::CuArray{Fl\
oat32,4}) at /home/deniz/.julia/dev/CUDA/lib/utils/call.jl:93                                                                                                                
 [4] cudnnAddTensor(::CuArray{Float32,4}, ::CuArray{Float32,4}; alpha::Int64, beta::Int64) at /home/deniz/.julia/dev/CUDA/lib/cudnn/tensor.jl:30                             
 [5] cudnnAddTensor at /home/deniz/.julia/dev/CUDA/lib/cudnn/tensor.jl:30 [inlined]                                                                                          
 [6] broadcasted(::typeof(+), ::CuArray{Float32,4}, ::CuArray{Float32,4}) at /home/deniz/.julia/dev/CUDA/lib/cudnn/nnlib.jl:210                                              
 [7] top-level scope at none:1                                                                                                                                               
                                                                                                                                                                             
julia> b .+ a;                                                                                                                                                               
ERROR: CUDNNError: CUDNN_STATUS_BAD_PARAM (code 3)                                                                                                                           
Stacktrace:                                                                                                                                                                  
 [1] throw_api_error(::CUDA.CUDNN.cudnnStatus_t) at /home/deniz/.julia/dev/CUDA/lib/cudnn/error.jl:19                                                                        
 [2] macro expansion at /home/deniz/.julia/dev/CUDA/lib/cudnn/error.jl:30 [inlined]                                                                                          
 [3] cudnnAddTensor(::Ptr{Nothing}, ::Base.RefValue{Float32}, ::CUDA.CUDNN.TensorDesc, ::CuArray{Float32,4}, ::Base.RefValue{Float32}, ::CUDA.CUDNN.TensorDesc, ::CuArray{Fl\
oat32,4}) at /home/deniz/.julia/dev/CUDA/lib/utils/call.jl:93                                                                                                                
 [4] cudnnAddTensor(::CuArray{Float32,4}, ::CuArray{Float32,4}; alpha::Int64, beta::Int64) at /home/deniz/.julia/dev/CUDA/lib/cudnn/tensor.jl:30                             
 [5] cudnnAddTensor at /home/deniz/.julia/dev/CUDA/lib/cudnn/tensor.jl:30 [inlined]                                                                                          
 [6] broadcasted(::typeof(+), ::CuArray{Float32,4}, ::CuArray{Float32,4}) at /home/deniz/.julia/dev/CUDA/lib/cudnn/nnlib.jl:210                                              
 [7] top-level scope at none:1                                                                                                                                               
```

gartangh:
@denizyuret , for a moment, I forgot that dot broadcasting also works for arrays with different dimensions.
In the commit above I fixed the issue.
When calling `a .+ b` or `b  .+ a`, we will now spawn a regular Julia Broadcast kernel, as before.
Inside `conv_bias_act!`, we now call `add_bias`, which will perform the in-place addition of the 2 tensors.
Please note that this works for `add_bias(x,b)`, but not for `add_bias(b,x)`, for the same reason that `a.=a.+b` works, but `b.=a.+b` does not (assuming the dimensions you provided above).

gartangh:
>Something is not quite right with cudnnAddTensor

@denizyuret , I have made some adjustments to the code in https://github.com/JuliaGPU/CUDA.jl/pull/321/commits/233ebe6fddd870e4c3747b43d029ea59968961fd.
This should make sure that `cudnnAddTensor` is only called when the inputs have the right dimensions.
For all other cases, Julia broadcasting is used as before.
```Julia
julia> x = CUDA.zeros(2,3,4,5);
julia> b = CUDA.ones(1,1,4,1);
julia> CUDA.CUDNN.add_bias(x,b);
julia> x
2×3×4×5 CuArray{Float32,4}:
[:, :, 1, 1] =
 1.0  1.0  1.0
 1.0  1.0  1.0
 ...
```

denizyuret:
> > Something is not quite right with cudnnAddTensor
> 
> @denizyuret , I have made some adjustments to the code in [233ebe6](https://github.com/JuliaGPU/CUDA.jl/commit/233ebe6fddd870e4c3747b43d029ea59968961fd).
> This should make sure that `cudnnAddTensor` is only called when the inputs have the right dimensions.
> For all other cases, Julia broadcasting is used as before.

Interestingly the cudnn documentation suggests all Julia broadcastable arrays should work: "Each dimension of the bias tensor A must match the corresponding dimension of the destination tensor C or must be equal to 1. " -- we should file a bug report :)

gartangh:
> Interestingly the cudnn documentation suggests all Julia broadcastable arrays should work: "Each dimension of the bias tensor A must match the corresponding dimension of the destination tensor C or must be equal to 1. " -- we should file a bug report :)

@denizyuret , here we go: https://developer.nvidia.com/nvidia_bug/3084210.

JeffBezanson:
> I feel like Julia and its packages always favor accuracy over performance.

No, there is no such rule or "tradition". Packages should do whatever's best for them, and authors should make independent decisions about what is best.

DhairyaLGandhi:
This is indeed very promising,thank you!

I'm still interested in diving deeper in resnet,  is there a profile of the call graph with this branch?

Also, it seems that all activations are now constant time, can we verify this against CuDNN implementations elsewhere

DhairyaLGandhi:
Also, are there updates to the numbers posted in https://github.com/JuliaGPU/CUDA.jl/pull/321#issuecomment-665801028

gartangh:
>No, there is no such rule or "tradition". Packages should do whatever's best for them, and authors should make independent decisions about what is best.

@JeffBezanson , I am selecting the fast algorithms over the accurate ones by default now.

> Also, are there updates to the numbers posted in #321 (comment)

@DhairyaLGandhi , here you go: (I made some changes to my benchmark setup, so the numbers are not exactly the same as in https://github.com/JuliaGPU/CUDA.jl/pull/321#issuecomment-665801028). All numbers are still in milliseconds, and this is tested on an NVIDIA V100 16GB.

>I'm still interested in diving deeper in resnet, is there a profile of the call graph with this branch?

@DhairyaLGandhi , the problem with ResNet-50 is gone after the changes to my benchmark setup.

| Network  | Input size | CuArrays.jl #master (https://github.com/JuliaGPU/CuArrays.jl/commit/6ebe463d4a2ad991d5b95388af12ee57f6cc91f9) | CUDA.jl #master (https://github.com/JuliaGPU/CUDA.jl/commit/8c61a0556a5bae6096958302cadc6bd3435666b9) | CUDA.jl #conv_bias_act (https://github.com/gartangh/CUDA.jl/commit/a1e83ee50edd818d113ab69468a48d7a32f3a96c) |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| VGG19  | (224,224,3,64) | 308 | 267 | 127 |
| Darknet-19  | (224,224,3,128) | 155 | 114 | 67 |
| ResNet-50  | (224,224,3,64) | 412 | 193 | 83 |
| Darknet-53  | (256,256,3,64) | 238 | 187 | 95 |

| Network  | Input size | CuArrays.jl #grouped_convolutions (https://github.com/gartangh/CuArrays.jl/commit/99ed388a0444604f19848b115edf299f926c1a4b)  | CUDA.jl #grouped_conv (https://github.com/gartangh/CUDA.jl/commit/feb35728a250219160949000c4c11c478d0a54e3) |
| ------------- | ------------- | ------------- | ------------- |
| ResNeXt-50  | (224,224,3,32) | 245 | 45 |

denizyuret:
I'd like to run some benchmarks but could not find the right combination of CUDA, GPUArrays, NNlib etc. that works with your branch (checking out the masters of all the others did not work), @gartangh can you advise?

gartangh:
@denizyuret 

CUDA v1.2.0 (#conv_bias_act from https://github.com/gartangh/CUDA.jl)
Flux v0.11.0 (#conv_bias_act from https://github.com/gartangh/Flux.jl)
GPUArrays v4.0.1 (#56bc004 from https://github.com/JuliaGPU/GPUArrays.jl)
Metalhead v0.5.0 (#1d9db5e from https://github.com/FluxML/Metalhead.jl)
NNlib v0.7.3 (#conv_bias_act from https://github.com/gartangh/NNlib.jl)

DhairyaLGandhi:
bump

maleadt:
bors try

DhairyaLGandhi:
Needs the different branch from NNlib,otherwise we wouldn't get the benefit I would imagine

maleadt:
> Needs the different branch from NNlib,otherwise we wouldn't get the benefit I would imagine

There's more than just the `conv_bias_act!` method here, e.g., algorithm selection, broadcast changes. Those have an effect already.

PR #431, opened 2020-09-16
CUDA quicksort

xaellison:
An implementation of quicksort to address: https://github.com/JuliaGPU/CUDA.jl/issues/93

The performance is solid, see `src/sorting/usage.jl` for quick performance tests. I intend to later include handling lists with a large number of duplicates, which currently can stymie the method for partitioning.

Hopefully the tests included can help clarify the inner workings. Be warned that this is likely not the most "Julian" code that's ever been written in Julia.

### DIFF
--- a/src/sorting.jl
+++ b/src/sorting.jl
@@ -134,7 +134,8 @@ end
 Each block evaluates `batch_partition` on consecutive regions of length blockDim().x
 from `lo` to `hi` of `values`.
 """
-function partition_batches_kernel(values::AbstractArray{T}, pivot, lo, hi, parity, lt::F1, by::F2) where {T,F1,F2}
+function partition_batches_kernel(values::AbstractArray{T}, pivot, lo, hi, parity, lt::F1,
+                                  by::F2) where {T,F1,F2}
     sums = @cuDynamicSharedMem(Int, blockDim().x)
     swap = @cuDynamicSharedMem(T, blockDim().x, sizeof(sums))
     batch_partition(values, pivot, swap, sums, lo, hi, parity, lt, by)
@@ -178,7 +179,8 @@ for storing the partition of each batch.
 
 Must only run on 1 SM.
 """
-@inline function consolidate_batch_partition(vals::AbstractArray{T}, pivot, lo, L, b_sums, parity, lt::F1, by::F2) where {T,F1,F2}
+@inline function consolidate_batch_partition(vals::AbstractArray{T}, pivot, lo, L, b_sums,
+                                             parity, lt::F1, by::F2) where {T,F1,F2}
     sync_threads()
     @inline N_b() = ceil(Int, L / blockDim().x)
     @inline batch(k) = threadIdx().x + k * blockDim().x
@@ -192,12 +194,20 @@ Must only run on 1 SM.
             if batch(my_iter) <= N_b()
                 seek_lo = lo + (batch(my_iter) - 1) * blockDim().x
                 seek_hi = lo + min(L, batch(my_iter) * blockDim().x)
-                b_sums[threadIdx().x] = seek_hi - find_partition(vals, pivot, seek_lo, seek_hi, parity, lt, by)
+                b_sums[threadIdx().x] =
+                    seek_hi - find_partition(vals, pivot, seek_lo, seek_hi, parity, lt, by)
             end
             my_iter += 1
         end
 
-        n_eff() = (batch_i != ceil(Int, L / blockDim().x) || L % blockDim().x == 0) ? blockDim().x : L % blockDim().x
+        function n_eff()
+            if batch_i != ceil(Int, L / blockDim().x) || L % blockDim().x == 0
+                blockDim().x
+            else
+                L % blockDim().x
+            end
+        end
+
         sync_threads()
         d = b_sums[batch_i - (my_iter - 1) * blockDim().x]
         c = n_eff() - d
@@ -220,6 +230,7 @@ Must only run on 1 SM.
         a += c
         b += d
     end
+
     sync_threads()
     return lo + a
 end
@@ -265,10 +276,13 @@ end
 """
 Launch batch partition kernel and sync
 """
-@inline function call_batch_partition(vals::AbstractArray{T}, pivot, swap, b_sums, lo, hi, parity, sync::Val{true}, lt::F1, by::F2) where {T, F1, F2}
+@inline function call_batch_partition(vals::AbstractArray{T}, pivot, swap, b_sums, lo, hi,
+                                      parity, sync::Val{true}, lt::F1, by::F2) where {T, F1, F2}
     L = hi - lo
     if threadIdx().x == 1
-        @cuda blocks=ceil(Int, L / blockDim().x) threads=blockDim().x dynamic=true shmem=blockDim().x*(sizeof(Int)+sizeof(T)) partition_batches_kernel(vals, pivot, lo, hi, parity, lt, by)
+        @cuda(blocks=ceil(Int, L / blockDim().x), threads=blockDim().x, dynamic=true,
+              shmem=blockDim().x*(sizeof(Int)+sizeof(T)),
+              partition_batches_kernel(vals, pivot, lo, hi, parity, lt, by))
         CUDA.device_synchronize()
     end
 end
@@ -276,7 +290,8 @@ end
 """
 Partition batches in a loop using a single block
 """
-@inline function call_batch_partition(vals::AbstractArray{T}, pivot, swap, b_sums, lo, hi, parity, sync::Val{false}, lt::F1, by::F2) where {T, F1, F2}
+@inline function call_batch_partition(vals::AbstractArray{T}, pivot, swap, b_sums, lo, hi,
+                                      parity, sync::Val{false}, lt::F1, by::F2) where {T, F1, F2}
     for temp in lo:blockDim().x:hi
         batch_partition(vals, pivot, swap, b_sums, temp, min(hi, temp + blockDim().x), parity, lt, by)
     end
@@ -305,7 +320,8 @@ the pivot hasn't changed and partition = `lo` or `hi`. If `stuck` reaches 2, rec
 it's possible that the first pivot will be that value, which could lead to an incorrectly
 early end to recursion if we started `stuck` at 0.
 """
-function qsort_kernel(vals::AbstractArray{T,N}, lo, hi, parity, sync::Val{S}, sync_depth, prev_pivot, lt::F1, by::F2, ::Val{dims}, stuck=-1) where {T, N, S, F1, F2, dims}
+function qsort_kernel(vals::AbstractArray{T,N}, lo, hi, parity, sync::Val{S}, sync_depth,
+                      prev_pivot, lt::F1, by::F2, ::Val{dims}, stuck=-1) where {T, N, S, F1, F2, dims}
     b_sums = @cuDynamicSharedMem(Int, blockDim().x)
     swap = @cuDynamicSharedMem(T, blockDim().x, sizeof(b_sums))
     shmem = sizeof(b_sums) + sizeof(swap)
@@ -350,9 +366,13 @@ function qsort_kernel(vals::AbstractArray{T,N}, lo, hi, parity, sync::Val{S}, sy
         if stuck < 2 && partition > lo
             s = CuDeviceStream()
             if sync_depth > 1
-                @cuda threads=blockDim().x dynamic=true stream=s shmem=shmem qsort_kernel(slice, lo, partition, !parity, Val(true), sync_depth - 1, pivot, lt, by, Val(1), stuck)
+                @cuda(threads=blockDim().x, dynamic=true, stream=s, shmem=shmem,
+                      qsort_kernel(slice, lo, partition, !parity, Val(true), sync_depth - 1,
+                      pivot, lt, by, Val(1), stuck))
             else
-                @cuda threads=blockDim().x dynamic=true stream=s shmem=shmem qsort_kernel(slice, lo, partition, !parity, Val(false), sync_depth - 1, pivot, lt, by, Val(1), stuck)
+                @cuda(threads=blockDim().x, dynamic=true, stream=s, shmem=shmem,
+                      qsort_kernel(slice, lo, partition, !parity, Val(false), sync_depth - 1,
+                      pivot, lt, by, Val(1), stuck))
             end
             CUDA.unsafe_destroy!(s)
         end
@@ -360,9 +380,13 @@ function qsort_kernel(vals::AbstractArray{T,N}, lo, hi, parity, sync::Val{S}, sy
         if stuck < 2 && partition < hi
             s = CuDeviceStream()
             if sync_depth > 1
-                @cuda threads=blockDim().x dynamic=true stream=s shmem=shmem qsort_kernel(slice, partition, hi, !parity, Val(true), sync_depth - 1, pivot, lt, by, Val(1), stuck)
+                @cuda(threads=blockDim().x, dynamic=true, stream=s, shmem=shmem,
+                      qsort_kernel(slice, partition, hi, !parity, Val(true), sync_depth - 1,
+                      pivot, lt, by, Val(1), stuck))
             else
-                @cuda threads=blockDim().x dynamic=true stream=s shmem=shmem qsort_kernel(slice, partition, hi, !parity, Val(false), sync_depth - 1, pivot, lt, by, Val(1), stuck)
+                @cuda(threads=blockDim().x, dynamic=true, stream=s, shmem=shmem,
+                      qsort_kernel(slice, partition, hi, !parity, Val(false), sync_depth - 1,
+                      pivot, lt, by, Val(1), stuck))
             end
             CUDA.unsafe_destroy!(s)
         end
@@ -372,18 +396,20 @@ function qsort_kernel(vals::AbstractArray{T,N}, lo, hi, parity, sync::Val{S}, sy
 end
 
 function quicksort!(c::AbstractArray{T,N}; lt::F1, by::F2, dims::Int) where {T,N,F1,F2}
-    MAX_DEPTH = CUDA.limit(CUDA.LIMIT_DEV_RUNTIME_SYNC_DEPTH)
+    max_depth = CUDA.limit(CUDA.LIMIT_DEV_RUNTIME_SYNC_DEPTH)
     len = size(c, dims)
 
+    1 <= dims <= N || throw(ArgumentError("dimension out of range"))
     otherdims = ntuple(i -> i == dims ? 1 : size(c, i), N)
 
-    kernel = @cuda launch=false qsort_kernel(c, 0, len, true, Val(MAX_DEPTH > 1), MAX_DEPTH, nothing, lt, by, Val(dims))
+    kernel = @cuda launch=false qsort_kernel(c, 0, len, true, Val(max_depth > 1),
+                                             max_depth, nothing, lt, by, Val(dims))
 
     get_shmem(threads) = threads * (sizeof(Int) + sizeof(T))
     config = launch_configuration(kernel.fun, shmem=threads->get_shmem(threads))
     threads = prevpow(2, config.threads)
 
-    kernel(c, 0, len, true, Val(MAX_DEPTH > 1), MAX_DEPTH, nothing, lt, by, Val(dims);
+    kernel(c, 0, len, true, Val(max_depth > 1), max_depth, nothing, lt, by, Val(dims);
            blocks=prod(otherdims), threads=threads, shmem=get_shmem(threads))
 
     return c
@@ -401,11 +427,6 @@ function Base.sort!(c::AnyCuArray; dims::Integer, lt=isless, by=identity, rev=fa
     if rev
         lt = !lt
     end
-
-    nd = ndims(c)
-    k = dims
-    sz = size(c)
-    1 <= k <= nd || throw(ArgumentError("dimension out of range"))
 
     quicksort!(c; lt, by, dims)
     return c
--- a/src/sorting.jl
+++ b/src/sorting.jl
@@ -109,14 +109,14 @@ Uses block index to decide which values to operate on.
 @inline function batch_partition(values, pivot, swap, sums, lo, hi, parity, lt::F1, by::F2) where {F1,F2}
     sync_threads()
     idx0 = Int(lo) + (blockIdx().x - 1) * blockDim().x + threadIdx().x
-    if idx0 <= hi
+    @inbounds if idx0 <= hi
          swap[threadIdx().x] = values[idx0]
          sums[threadIdx().x] = 1 & flex_lt(pivot, swap[threadIdx().x], parity, lt, by)
     else
-        @inbounds sums[threadIdx().x] = 1
+         sums[threadIdx().x] = 1
     end
     sync_threads()
-    val = merge_swap_shmem(swap[threadIdx().x], threadIdx().x, 1, sums, swap)
+    val = merge_swap_shmem(@inbounds(swap[threadIdx().x]), threadIdx().x, 1, sums, swap)
     temp = 2
     while temp < blockDim().x
         val = merge_swap_shmem(val, threadIdx().x, temp, sums, swap)
@@ -124,7 +124,7 @@ Uses block index to decide which values to operate on.
     end
     sync_threads()
 
-    if idx0 <= hi
+    @inbounds if idx0 <= hi
          values[idx0] = val
     end
     sync_threads()
@@ -203,8 +203,10 @@ Must only run on 1 SM.
         c = n_eff() - d
         to_move = min(b, c)
         sync_threads()
-        if threadIdx().x <= to_move
-            swap = vals[lo + a + threadIdx().x]
+        swap = if threadIdx().x <= to_move
+            vals[lo + a + threadIdx().x]
+        else
+            zero(eltype(vals))  # unused value
         end
         sync_threads()
         if threadIdx().x <= to_move
@@ -240,8 +242,10 @@ elements spaced by `stride`. Good for sampling pivot values as well as short sor
         for level in 0:L
             # get left/right neighbor depending on even/odd level
             buddy = threadIdx().x - 1 + 2 * (1 & (threadIdx().x % 2 != level % 2))
-            if 1 <= buddy <= L && threadIdx().x <= L
-                buddy_val = swap[buddy]
+            buddy_val = if 1 <= buddy <= L && threadIdx().x <= L
+                 swap[buddy]
+            else
+                zero(eltype(swap)) # unused value
             end
             sync_threads()
             if 1 <= buddy <= L && threadIdx().x <= L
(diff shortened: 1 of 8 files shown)


maleadt:
Nice, thanks! I'll try to give this a proper look, but two quick questions:
- not all devices can launch 1024 threads, is it possible to generalize that?
- can you use this to implement `Base.sort!`?

xaellison:
Interestingly, it seems to speed up when the block size is halved. The trade-offs when you use 512 threads compared to 1024:
- Bubble sort is half as many iterations
- The second half of partitioning goes down to 1 block, which means it should need ~twice as many steps. Later in the recursion, these can run in parallel streams.
- The size of static shared memory (per block) is cut in half. Even if there are twice as many blocks would they be easier to schedule because they're smaller?
[Edit]
- Also, each grid which "batch partitions" could waste n-1 threads for a grid with blocks of size n. 

Using `src/sorting/usage.jl`:
```
julia> speed_test_random(Int32, Int(1e7), 1024)
BenchmarkTools.Trial:
  memory estimate:  1.69 KiB
  allocs estimate:  58
  --------------
  minimum time:     684.429 ms (0.00% GC)
  median time:      692.137 ms (0.00% GC)
  mean time:        691.690 ms (0.00% GC)
  maximum time:     698.537 ms (0.00% GC)
  --------------
  samples:          10
  evals/sample:     1

julia> speed_test_random(Int32, Int(1e7), 512)
BenchmarkTools.Trial:
  memory estimate:  1.69 KiB
  allocs estimate:  58
  --------------
  minimum time:     630.825 ms (0.00% GC)
  median time:      642.350 ms (0.00% GC)
  mean time:        641.250 ms (0.00% GC)
  maximum time:     645.002 ms (0.00% GC)
  --------------
  samples:          10
  evals/sample:     1
```

maleadt:
> Interestingly, it seems to speed up when the block size is halved.

I've noticed the same in https://github.com/JuliaGPU/GPUArrays.jl/pull/301#issuecomment-659453252, maximizing occupancy doesn't always guarantee best performance.

maleadt:
This one looks related:
```
      From worker 2:	ERROR: LoadError: There was an error during testing
      From worker 2:	in expression starting at /var/lib/buildkite-agent/builds/3-cyclops-juliacomputing-io/julialang/cuda-dot-jl/examples/quicksort.jl:35
      From worker 2:	error in running finalizer: CUDA.CuError(code=CUDA.cudaError_enum(0x000002bc), meta=nothing)
```
```
julia> CUDA.cudaError_enum(0x000002bc)
CUDA_ERROR_ILLEGAL_ADDRESS::cudaError_enum = 0x000002bc
```

As well as:
```
Error in testset quicksort:
  LoadError: CUDA error: too many resources requested for launch (code 701, ERROR_LAUNCH_OUT_OF_RESOURCES)
```

examples/quicksort.jl
```diff
@@ -0,0 +1,53 @@
...
+    c = CuArray(a)
+    sort!(c)
+    @test c == sort(a)
+    c = CuArray(a)
+    CUDA.quicksort!(c, half_size)
+    @test c == sort(a)
+end
+
+
+# test pre-sorted
```
maleadt: I don't think these warrant a new `examples` file (which are fairly expensive to test, too). Just put these in `test/sorting`.jl?

src/sorting/quicksort.jl
```diff
@@ -0,0 +1,416 @@
+"""
```
maleadt: Let's just call this `src/sorting.jl` for now (unless you were planning to develop other implementation? :slightly_smiling_face:)

src/sorting/quicksort.jl
```diff
@@ -0,0 +1,416 @@
...
+    if i == 1
+        if L <= N_t ^ 2
+            # perform batch partition and do not store partition indices
+            @cuda blocks=N_b threads=N_t dynamic=true partition_batches_kernel(vals, pivot, lo, L, block_dim, parity)
+        else
+            # store batch partition indices in global array
+            @cuda blocks=N_b threads=N_t dynamic=true partition_batches_kernel(vals, pivot, global_b_sums, lo, L, block_dim, parity)
+        end
+    end
+    CUDA.device_synchronize()
```
maleadt: You shouldn't need  prefixes, since this code will be part of the CUDA module.

src/sorting/quicksort.jl
```diff
@@ -0,0 +1,416 @@
...
+    - Shared memory
+This assumes that register usage doubles with elevated debug level, but usage of
+other resources is unaffected.
+"""
+function quicksort_block_size(T)
+    # requirements
+    # this is a guess for older architectures. 64 is fine for RTX 2070
+    registers_per_thread = 128
+    if Base.JLOptions().debug_level >= 2
+        registers_per_thread *= 2
```
maleadt: That's pretty inaccurate. You can always inspect the register usage of a compiled kernel. But even better, can't you use the occupancy API and skip inspecting device attributes below, too?

src/sorting/usage.jl
```diff
@@ -0,0 +1,20 @@
+using BenchmarkTools
```
maleadt: This shouldn't be part of CUDA.jl. You can add it to the `perf/` benchmarks too, although I would there test the invocation of `Base.sort`, not `CUDA.quicksort` specifically.

test/quicksort.jl
```diff
@@ -0,0 +1,126 @@
...
+Test Passed
+
+julia>
+```
+"""
+
+using DataStructures, Random, Test
+
+#using CUDA
+include("../src/sorting/quicksort.jl")
```
maleadt: CUDA.jl should just include this functionality unconditionally. I assume you left it out for easier development? With Revise that shouldn't be necessary though.

test/quicksort.jl
```diff
@@ -0,0 +1,126 @@
...
+
+full_size = quicksort_block_size(Float32)
+test_consolidate_partition(Float32, 10000, full_size ÷ 4, 0, 10000, 0)
+test_consolidate_partition(Float32, 10000, full_size ÷ 2, 0, 10000, 0)
+test_consolidate_partition(Float32, 10000, full_size, 0, 10000, 0)
+
+full_size = quicksort_block_size(Int64)
+test_consolidate_partition(Int64, 10000, full_size ÷ 4, 0, 10000, 0)
+test_consolidate_partition(Int64, 10000, full_size ÷ 2, 0, 10000, 0)
+test_consolidate_partition(Int64, 10000, full_size, 0, 10000, 0)
```
maleadt: Maybe put all these tests in a `@testset "quicksort"` and also do some generic testing of `Base.sort` and `Base.sort!`, which is what most users will end up using.

src/sorting/quicksort.jl
```diff
@@ -0,0 +1,416 @@
...
+
+    # this is safe assuming a 1d block is always executed on 1 multiprocessor
+    block_size = 1
+    while all(block_size .* (1, registers_per_thread, shmem_per_thread) .< (MAX_BLOCK, MAX_REG, MAX_SHMEM))
+        block_size *= 2
+    end
+    return block_size
+end
+
+function Base.sort!(c :: CuArray{T}) where T
```
maleadt: `Base.sort!` has quite some kwargs, can some of those be implemented already by your `quicksort!` method?

maleadt (changes requested):
Starting to look very good, thanks! :-)

src/sorting/quicksort.jl
xaellison: I didn't know that functionality existed - that's awesome! I had just measured register usage with nvvp locally. I see how I could set a `config` option when I call `@cuda`, and that could set threads based on register usage (or via the functions in `occupancy.jl`). There is a slight problem with that. `qsort_kernel` sets static shmem using a Val type, and that Val matches block size. I don't see an easy way to use the `config` option unless I switch to dynamic shmem, which seems unnecessary.

One way I could see getting around this is creating a macro which finds the largest block size possible (out of [256, 512, 1024, 2048]). This could be done by calling creating a `HostKernel` for each block size and using the occupancy API to pick the largest viable option. Each `HostKernel` would have the corresponding Val argument. Does that seem like a reasonable approach or is there something easier?

xaellison:
@maleadt  thanks for the feedback! I'll happily implement all the code-organization/testing ones in the next push. The ones closer to the actual sorting will take some thought

src/sorting/quicksort.jl
maleadt: > `qsort_kernel` sets static shmem using a Val type, and that Val matches block size. I don't see an easy way to use the `config` option unless I switch to dynamic shmem, which seems unnecessary.

Why don't you want to use dynamic shared memory? It's exactly meant for this use case. Otherwise you'd compile multiple kernels, and the occupancy API supports using a closure that determines the amount of dynamic shared memory required for this configuration.

xaellison: The original reason was performance. There was something like a factor of two difference for `partition_batches_kernel` (as profiled by nvvp).  I never committed that code unfortunately. Assuming performance is significantly better, having a few extra methods sounds like a worthwhile cost, no?

The real, current reason is if I switch to using dynamic memory the code breaks. `partition_batches_kernel` seems to get confused because it uses two arrays in shared memory. When using static shmem, it works fine.

Either this is a bug, or I misunderstood how to allocate dynamic shared memory. I've taken the risk of filing an issue :)
https://github.com/JuliaGPU/CUDA.jl/issues/555

maleadt: > The original reason was performance.

I don't understand this; there shouldn't be a difference, unless you use different amounts of memory. Are you sure your `shmem` calculation was correct? It should be in bytes, not in number of elements.

src/sorting/quicksort.jl
xaellison: Some certainly can. `dims` is easy by using views. `rev` is possible with minor modifications. These two I plan to include in this PR.

Others are harder. `by` could be done in two ways:
1. Custom compiled kernels.
2. Broadcasting `by` to an array which would be our keys, and then sort has to move twice as much data (keys and values).
Of the two, I prefer 1. Either way, this probably has to be a follow-up feature.

`lt` could be defined similarly, but it could only be done if there is always a corresponding `==`. This is because the sort alternates between partitioning on `<` and `<=`.

The other two args, `alg` and `order` do not seem applicable. I can't even find a good example of what `order` does.

src/sorting/quicksort.jl
xaellison: Ok - thanks for pointing out the offset logic. I can't seem to reproduce the performance gap which is good. The overall sort runs in about the same amount of time. Interestingly there is a big jump in register usage: 48 -> 72.

src/sorting.jl
```diff
@@ -390,5 +390,5 @@ function quicksort!(c :: CuArray{T}) where T
     end
 
     b_sums = CuArray{Int32}(undef, ceil_div(N, 32))
-    CUDA.@sync(@cuda config=get_config  qsort_kernel(c, Int32(0), N, b_sums, Val(true), Int8(1)))
+    CUDA.@sync(@cuda threads=64 shmem=(sizeof(T) + 4)*64  qsort_kernel(c, Int32(0), N, b_sums, Val(true), Int8(1)))
```
maleadt: See here for an example on using the occupancy API with a shmem callback: https://github.com/JuliaGPU/CUDA.jl/blob/bc0184eaba90c303e73c44a1768b6b7cb36ed669/src/mapreduce.jl#L196-L197

maleadt:
I guess something went wrong here merging? Better to rebase anyway.

xaellison:
> I guess something went wrong here merging? Better to rebase anyway.

That commit should be the result of a rebase. It looks like tests passed on build 303 and it failed because of missing NetworkOptions.

maleadt:
Well, something still went wrong because there's a bunch of commits of mine in here: https://github.com/JuliaGPU/CUDA.jl/pull/431/commits

xaellison:
How do you feel about me closing this PR and opening a new one :)

maleadt:
Just `git reset --soft origin/master` and then commit everything afresh and `push -f` it here.

xaellison:
Thank you - obviously I'm horribly out of practice with git

src/sorting.jl
```diff
@@ -0,0 +1,392 @@
...
+    out
+end
+
+function pow2_floor(x)
+    out = 1
+    while out * 2 <= x
+        out *= 2
+    end
+    out
+end
```
xiaodaigh: i don't know about numerical math like this but I assume this is faster than `floor(log(2, x))`
Seems to be not much difference in speed

```
function pow2_ceil(x)
    out = 1
    while out < x
        out *= 2
    end
    out
end

function pow2_ceil_shift(x)
    out = 1
    while out < x
        out <<= 1
    end
    out
end

using BenchmarkTools

function pow2_ceil_v2(x)
    2^ceil(Int, log(2, x))
end


x = abs(Float32(rand(Int16)))

pow2_ceil(x) == pow2_ceil_shift(x)
pow2_ceil(x) == pow2_ceil_v2(x)

@benchmark pow2_ceil($x)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     25.074 ns (0.00% GC)
  median time:      25.276 ns (0.00% GC)
  mean time:        25.721 ns (0.00% GC)
  maximum time:     81.946 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     997

@benchmark pow2_ceil_shift($x)

BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     25.050 ns (0.00% GC)
  median time:      25.251 ns (0.00% GC)
  mean time:        25.636 ns (0.00% GC)
  maximum time:     81.864 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     998


@benchmark pow2_ceil_v2($x)

BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     19.038 ns (0.00% GC)
  median time:      20.542 ns (0.00% GC)
  mean time:        21.411 ns (0.00% GC)
  maximum time:     200.600 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     998
```

xiaodaigh:
Can we compare the speed to the SortingAlgorithms.jl's Radix sort performance wise?

src/sorting.jl
Ellipse0934: `ceil_div` can be replaced by `cld`.

You can use this code as replacements to avoid iterative algorithms.
```julia
@inline function pow2_floor_gpu(x)
    CUDA.popc(x) != 1 && return 1 << (sizeof(x)*8 - CUDA.ffs(CUDA.brev(x)))
    return x
end

@inline function pow2_ceil_gpu(x)
    CUDA.popc(x) != 1 && return 1 << (sizeof(x)*8 - CUDA.ffs(CUDA.brev(x)) + 1)
    return x
end
```

xaellison:
The update for Julia 1.6 forced me to make some major updates, which have made this work a lot better (as a standalone algorithm). I think this is up to date with master, so I'm not sure why this is failing in CI on a dependency issue.

That said, I am getting an error when I try to call `quicksort!` after including `sorting.jl` in `src/CUDA.jl`:
```
julia> quicksort!(c)
Internal error: encountered unexpected error during compilation of #launch_configuration#22:
LLVM.LLVMException(info="Symbols not found: [ __nv_llmax ]
")
handle_error at /home/ec2-user/.julia/packages/LLVM/dVU7J/src/core/context.jl:105
```


Because of this, I have *not* yet addressed all the code organization comments above.

xaellison:
> Can we compare the speed to the SortingAlgorithms.jl's Radix sort performance wise?

Sure! This is from an aws ec2 instance of type p3.2xlarge. Radix sort's performance varies significantly with type, but Int32 gives you a roughly average comparison. This validates that the sort is correct, too.
```
julia> CUDA.limit(CUDA.LIMIT_DEV_RUNTIME_SYNC_DEPTH)
4

julia> versioninfo()
Julia Version 1.6.0-DEV.1597
Commit 059ea247b0 (2020-11-30 12:49 UTC)
Platform Info:
  CPU: Intel(R) Xeon(R) CPU E5-2686 v4 @ 2.30GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.0 (ORCJIT, broadwell)
Environment:
  JULIA_CUDA_USE_BINARYBUILDER = false

julia> CUDA.versioninfo()
CUDA toolkit 11.0.3, local installation
CUDA driver 11.0.0
NVIDIA driver 450.80.2

Libraries: 
- CUBLAS: 11.2.0
- CURAND: 10.2.1
- CUFFT: 10.2.1
- CUSOLVER: 10.6.0
- CUSPARSE: 11.1.1
- CUPTI: 13.0.0
- NVML: 11.0.0+450.80.2
- CUDNN: 8.0.4 (for CUDA 11.0.0)
- CUTENSOR: missing

Toolchain:
- Julia: 1.6.0-DEV.1597
- LLVM: 11.0.0
- PTX ISA support: 3.2, 4.0, 4.1, 4.2, 4.3, 5.0, 6.0, 6.1, 6.3, 6.4, 6.5, 7.0
- Device support: sm_35, sm_37, sm_50, sm_52, sm_53, sm_60, sm_61, sm_62, sm_70, sm_72, sm_75, sm_80

Environment:
- JULIA_CUDA_USE_BINARYBUILDER: false

1 device:
  0: Tesla V100-SXM2-16GB (sm_70, 11.473 GiB / 15.782 GiB available)

julia> @benchmark quicksort!(c) setup=(a = rand(Int32, Int(1e7)); c=CuArray(a)) teardown=(@assert sort(a)==Array(c)) samples=10 evals=1 seconds=20
BenchmarkTools.Trial: 
  memory estimate:  2.23 KiB
  allocs estimate:  53
  --------------
  minimum time:     298.225 ms (0.00% GC)
  median time:      307.526 ms (0.00% GC)
  mean time:        307.692 ms (0.00% GC)
  maximum time:     319.580 ms (0.00% GC)
  --------------
  samples:          10
  evals/sample:     1

julia> using SortingAlgorithms

julia> @benchmark sort!(a, alg=RadixSort) setup=(a = rand(Int32, Int(1e7))) teardown=(@assert issorted(a)) samples=10 evals=1 seconds=20
BenchmarkTools.Trial: 
  memory estimate:  38.24 MiB
  allocs estimate:  10
  --------------
  minimum time:     473.968 ms (0.00% GC)
  median time:      484.322 ms (0.81% GC)
  mean time:        512.827 ms (6.67% GC)
  maximum time:     646.689 ms (24.90% GC)
  --------------
  samples:          10
  evals/sample:     1

julia> 
```

maleadt:
I missed your comments -- are you still running into the `__nv_llmax` issue?

xaellison:
> I missed your comments -- are you still running into the `__nv_llmax` issue?

Putting all of `sorting.jl` into its own module fixed that

maleadt:
Disabling memcheck is only for incompatibilities, while this seems like a legitimate issue:

```


      From worker 2:	========= CUDA-MEMCHECK
      From worker 2:	========= Invalid __global__ write of size 8
      From worker 2:	=========     at 0x0000fdf0 in _Z30julia_qsort_async_kernel_3933713CuDeviceArrayI4Int8Li1ELi1EE5Int32S1_3ValILitrueEE
      From worker 2:	=========     by thread (0,0,0) in block (0,0,0)
      From worker 2:	=========     Address 0x7ffed6a08c28 is out of bounds
```

xaellison:
I have now reproduced the memcheck error. It occurs when sorting a [list with many many duplicates](https://github.com/JuliaGPU/CUDA.jl/pull/431/files#diff-8a00fd5468cbab4c89477359a95159628ab1a84e095b2c1008c218a58e46309aR198). However, if I don't run the test through memcheck, the same test passes (not just runs, but runs correctly).

The recursion can go pretty deep for this case. Memcheck makes all launches [blocking](https://docs.nvidia.com/cuda/cuda-memcheck/index.html#known-issues). If I prevent launching kernels at depth > 24, the error goes away. That's the maximum sync depth. The same error occurs if we run the following MWE through memcheck:

```
using CUDA

function kernel(n)
    if n > 0
        @cuda dynamic=true kernel(n - 1)
    end
    return nothing
end

@cuda kernel(24)
synchronize()
@info "no problem"
@cuda kernel(24)
synchronize()
@info "no problem"
# this fails:
@cuda kernel(25)
synchronize()
```

To prevent such launches from happening, I can make the kernel aware of when it has found a large section of identical values.

maleadt:
> If I prevent launching kernels at depth > 24, the error goes away. That's the maximum sync depth.

So it seems important to honor this limit.

xaellison:
> 
> 
> > If I prevent launching kernels at depth > 24, the error goes away. That's the maximum sync depth.
> 
> So it seems important to honor this limit.

Agreed. This has been taken care of. The recursion now is aware of the maximum sync depth available, and switches to unsynchronized launches at that depth.

Additionally, keywords `rev` and `dims` are implemented for `Base.sort` and `Base.sort!`. `by` is implemented for `Base.sort`.

maleadt:
I rebased the PR to get CI working again.

xaellison:
Thank you for fixing CI @maleadt. I've looked at the most recent [failure](https://buildkite.com/julialang/cuda-dot-jl/builds/487). There is one error and it only fails when all of the following are true:
1. Compute capability <= 6.x. (quadro p5000, p6000). Pool type doesn't affect anything.
2. Calling `sort` with `by != identity`. `by=x->x` will fail.
3. Multidimensional array with `dims` (even if size is effectively 1D like `(1, 10000)`).
4. `test/sorting.jl` is launched via `test/runtests.jl`. If I manually `include` `test/sorting.jl` from repl, everything passes.

When I inspect the result of the sort, it is clear that it has been correctly partitioned once. That makes me suspect that synchronization has been affected. Is there any obvious reason why that might be the case?

maleadt:
Another rebase; also did some minor clean-up and updates for new APIs.

I removed the comment about using a custom stream, as I don't think library functions should use their own. I hope to introduce a task-local stream at some point, which would be used automatically by all called CUDA functions.

I couldn't reproduce the synchronization function, even on older GPUs. I was wondering though, can't we pass the `by` function to `quicksort` and call it lazily instead of eagerly mapping? The current approach is very costly, memory wise, and we can't do `sort!(; by=...)`.

maleadt (approved):

maleadt:
> I was wondering though, can't we pass the `by` function to `quicksort` and call it lazily instead of eagerly mapping? The current approach is very costly, memory wise, and we can't do `sort!(; by=...)`.

I've pushed some commits doing so for `lt` and `by`. At a very quick glance, 10% perf improvement for `by`, and 25% using `by` and `rev`.

src/sorting.jl
```diff
@@ -360,7 +360,7 @@ function quicksort!(c::AbstractArray{T}; lt::F1, by::F2) where {T,F1,F2}
 
     kernel = @cuda launch=false qsort_kernel(c, 0, N, true, Val(MAX_DEPTH > 1), MAX_DEPTH, nothing, lt, by)
 
-    get_shmem(threads) = threads * (sizeof(Int32) + max(4, sizeof(T)))
```
maleadt: @xaellison why did you clamp to `4` here? The other `shmem` calculations (in the kernel) didn't do that.

xaellison: Off the top of my head, I can't remember, but that seems wrong. `get_shmem(threads) = threads * (sizeof(Int32) + sizeof(T))` would be consistent

src/sorting.jl
```diff
@@ -102,13 +102,13 @@ Partition the region of `values` after index `lo` up to (inclusive) `hi` with
 respect to `pivot`. This is done by a 'binary' merge sort, where each the values
 are sorted by a boolean key: how they compare to `pivot`. The comparison is
 affected by `parity`. See `flex_lt`. `swap` is an array for exchanging values
-and `sums` is an array of Int32s used during the merge sort.
```
maleadt: @xaellison did you have a specific reason for using Int32? I didn't notice performance degradation by switching to Int (occupancy was identical despite the increased shared memory usage), and it avoids exceptions paths due to Int<->Int32 conversions.

xaellison: That was just for reducing shmem usage. If the overall performance isn't worse, Int makes sense instead of Int32.

ViralBShah:
Just found this today. I'm so excited to see this. Welcome @xaellison

xaellison:
> Just found this today. I'm so excited to see this. Welcome @xaellison

xaellison:
That close was unintentional

xaellison:
> Just found this today. I'm so excited to see this. Welcome @xaellison

Thank you, @ViralBShah . I'm really excited to share this contribution.

maleadt:
Last thing I wanted to improve, for now: In https://github.com/JuliaGPU/CUDA.jl/pull/431/commits/14cbde4cacfa72f0e331573dd953d5bacd9ed2e2 I changed how `dims` works, instead of calling `sort` multiple times the kernel now indexes the array using the block index. The dynamic kernel launches are still separate though, but we can't merge those because of the data-dependent execution of quicksort. Still, this massively improves performance of ND-sort, e.g. for a 1024x1024 array sorting takes 20ms instead of 700.

maleadt:
Some quick benchmarking: my Quadro RTX 5000 outperforms my heavily-overclocked 6600K at about 65536 elements. In the ND case, quicksort can be parallelized across dimensions and the performance is much better.

image

image

Times in ns, both axes log scale.

Comparing to the times in https://github.com/ihonen/cuda-sort, Float32 sort takes 7s vs. 2s in Thrust on a much less powerful GPU. So that's not great, but a good starting point :slightly_smiling_face:

xaellison:
@maleadt are you using the default sync depth = 2?  Sync depth >= 4 usually reduces runtime by  25%, and up to 40%. This was run on a V100. It doesn't catch up to Thrust, but it is better.
image


Data collected with lines like:
`@benchmark begin sort!(c); synchronize() end setup=(a = rand(Float32, Int(1e7)); c=CuArray(a)) teardown=(@assert sort(a)==Array(c)) samples=10 evals=1 seconds=20`

maleadt:
Ah interesting, I guess we should then bump it to the (still reasonable) value of 4, I guess? However, from the CUDA handbook:

> The limit cudaLimitDevRuntimeSyncDepth is especially important, since each nesting level costs up to 150MB of device memory.

And from the latest docs:

> The majority of reserved memory is allocated as backing-store for parent kernel state, for use when synchronizing on a child launch. Conservatively, this memory must support storing of state for the maximum number of live threads possible on the device. This means that each parent generation at which cudaDeviceSynchronize() is callable may require up to 860MB of device memory, depending on the device configuration, which will be unavailable for program use even if it is not all consumed. 

So probably too costly to bump globally...

maleadt:
> I've looked at the most recent [failure](https://buildkite.com/julialang/cuda-dot-jl/builds/487). There is one error and it only fails when all of the following are true:
> 
>     1. Compute capability <= 6.x. (quadro p5000, p6000). Pool type doesn't affect anything.
> 
>     2. Calling `sort` with `by != identity`. `by=x->x` will fail.
> 
>     3. Multidimensional array with `dims` (even if size is effectively 1D like `(1, 10000)`).
> 
>     4. `test/sorting.jl` is launched via `test/runtests.jl`. If I manually `include` `test/sorting.jl` from repl, everything passes.
> 
> 
> When I inspect the result of the sort, it is clear that it has been correctly partitioned once. That makes me suspect that synchronization has been affected. Is there any obvious reason why that might be the case?

Ah, we're still running into this apparently. Now it happened on an RTX2080, so it doesn't seem compute capability-related. The issue seems to require running under `--check-bounds=yes`, whereas running under `cuda-memcheck` fails to reproduce the issue. I thought it would be a failed sync due to exceeding the max sync depth, but adding error checks didn't reveal anything...

maleadt:
Oh FFS this is another case of https://github.com/JuliaGPU/CUDAnative.jl/issues/4 :sob: Forcing a `trap` instead of `exit`, https://github.com/JuliaGPU/GPUCompiler.jl/blob/1b436f83a7e8e89d9588afd52fd4dc168a8e5a2f/src/ptx.jl#L288-L294, fixes the issue which I reduced to:

```julia
using CUDA, Test

function main()
    for i in 1:typemax(Int)
        @show i
        A = rand(1:10, (2, 100000))
        d_A = CuArray(A)
        B = sort(A; dims=2)
        d_B = sort(d_A; dims=2)
        for x in (B, Array(d_B))
            @test issorted(x[1,:])
            @test issorted(x[2,:])
        end
        @test B == Array(d_B)
    end
end

isinteractive() || main()
```

(but which only reproduces on select hardware/driver combinations, here an RTX 2080 Ti on driver 450.80.2)

Filed as NVIDIA bug #3231266

maleadt:
All green! Let's merge this :rocket:
@xaellison Thanks for your work here!

PR #642, opened 2021-01-12
README: add some information on viewing capabilities of your devices

DilumAluthge:
I included both `CUDA.capability` and `CUDA.versioninfo`. This is because I think the output of `CUDA.capability` (e.g. `6.0`) is easier for a beginner to interpret than the e.g. `sm_60` output by `CUDA.versioninfo`.


DilumAluthge:
Thanks! Looks great.

PR #806, opened 2021-04-03
fix bounds check for reverse kwarg and add tests

kshyatt:
why not just use `ndims`? Also this wasn't throwing earlier and now does without progressing to the underscored methods.


maleadt (approved):
LGTM, thanks!

PR #820, opened 2021-04-09
Make reduce_warp exception free

vchuravy:


vchuravy:
I also noticed that reduce_block has a call to `fldmod1` that is not inlined and which itself can throw an exception.

```
function fld1(x::T, y::T) where T<:Integer
    d = Core.Intrinsics.sdiv_int(x, y)
    return d + (!signbit(x ⊻ y) & (d * y != x))
end
fldmod1(x, y) = (fld1(x, y), mod1(x, y))
```

Avoids the checked division.

maleadt (approved):
LGTM.

Time for a device override of `fldmod1`? Or should we tune the Base compiler to inline a bit more aggressively so that functions like this are inlined?

PR #823, opened 2021-04-11
sort bugfix

### LINKED ISSUE #822: Sort overwriting values in target array
**Describe the bug**

`sort`/`sort!` will return an array which isn't a permutation of the input for certain `by` arguments. Specifically, it will only be noticeable for `by` functions where multiple elements have the same output.

**To reproduce**

```julia
a = collect(1:200)
@info "$(length(Set(a))) distinct vals"
c = CuArray(a)
sort!(c; by=x->x % 4)
synchronize()
@info "For x % 4, $(length(Set(Array(c)))) distinct vals"

c = CuArray(a)
sort!(c; by=x->x % 3)
synchronize()
@info "For x % 3, $(length(Set(Array(c)))) distinct vals"

c = CuArray(a)
sort!(c; by=x->x % 2)
synchronize()
@info "For x % 2, $(length(Set(Array(c)))) distinct vals"
```
...
```
[ Info: 200 distinct vals
[ Info: For x % 4, 4 distinct vals
[ Info: For x % 3, 3 distinct vals
[ Info: For x % 2, 2 distinct vals
```

</p>
</details>

**Expected behavior**
The `by` argument should not change how many distinct values are in the list being sorted.

**Additional context**

The problem is in `bubble_sort`. At each step, it currently decides whether a thread should swap with its neighbor based on the following condition: `(threadIdx().x < buddy) != flex_lt(swap[threadIdx().x], buddy_val, false, lt, by)`
Obviously `(threadIdx().x < buddy)` is true for exactly one of each swapping pair. If `by` is the same for two adjacent values, then `flex_lt(swap[threadIdx().x], buddy_val, false, lt, by)` will be the same for _both_ threads in a swapping pair (since handling equality is hardcoded in as `false`). That means exactly one thread will decide to overwrite its value with the other's. This effect is unnoticeable for monotonic `by`, because the overwrite is with an identical value.

The above example shows this is limited to `bubble_sort` by using input smaller than the blocksize.

This is not limited to any specific hardware or version.

xaellison:
Fix #822


xaellison:
tried including a performance thing - it seems to need more testing

PR #847, opened 2021-04-19
Remove extraneous apostrophe

kshyatt:
"scenarios" is a perfectly valid English word.


maleadt (approved):
LGTM. Best to `[skip tests]` on doc-only fixes though.

PR #911, opened 2021-05-17
Just some missing tests for CUBLAS

kshyatt:


maleadt (approved):
LGMT. At some point we should try and get rid of some global state in these tests though, those global data arrays make it hard to reorder / parallelize testing, and are annoying to debug (since you can't easily isolate tests to run in the REPL).

PR #1042, opened 2021-07-09
restore lost tests

vchuravy:


vchuravy:
Seems to have gone missing in #908

simonbyrne:
Does it need to be `include`ed somewhere?

vchuravy:
> Does it need to be `include`ed somewhere?

https://github.com/JuliaGPU/CUDA.jl/blob/98874eee6e36ff6e0939a77c4a43096d3133643b/test/runtests.jl#L63 

We automatically pick them up.

The 1.8 failures are due to upstream Julia changes.

PR #1197, opened 2021-10-11
2-arg `show`

### LINKED ISSUE #1190: Should compact `show` mark `CuArray`s?
The compact printing used within a Tuple etc. doesn't at present distinguish a CuArray from an ordinary Array:
```julia
julia> (; x = cu([1f0]))  # CUDA v3.4.2
(x = Float32[1.0],)

julia> (; x = [1f0])
(x = Float32[1.0],)
```
Maybe it should? This came up in  https://github.com/FluxML/Zygote.jl/issues/1095 , but more generally Zygote likes to return tuples of arrays, and having a visible distinction might be helpful. 

This could just be `cu(Float32[1.0])` but perhaps that's not right for all eltypes.

mcabbott:
Closes #1190

Prints the Float32 case most compactly, although perhaps it would be less confusing to write out `cu(Float32[0.7821349, 0.99106485])` in the end? Certainly simpler.
```
julia> x = rand(2)
2-element Vector{Float64}:
 0.7821348955171186
 0.9910648595224902

julia> (f32 = cu(x), f64 = CuArray(x), i32 = CuArray(Int32[1,2,3]))
(f32 = cu([0.7821349, 0.99106485]), f64 = CuArray([0.7821348955171186, 0.9910648595224902]), i32 = cu(Int32[1, 2, 3]))
```

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -510,6 +510,22 @@ Adapt.adapt_storage(::CuArrayAdaptor{B}, xs::AbstractArray{T,N}) where {T<:Union
 @inline cu(xs; unified::Bool=false) = adapt(CuArrayAdaptor{unified ? Mem.UnifiedBuffer : Mem.DeviceBuffer}(), xs)
 Base.getindex(::typeof(cu), xs...) = CuArray([xs...])
 
+## show
+
+function Base.show(io::IO, A::AnyCuArray{T}) where {T}
+  if T <: Union{Float32, ComplexF32}
+    # then no need to write cu(Float32[1.0])
+    io = IOContext(io, :typeinfo => Vector{Float32})
+    print(io, "cu(")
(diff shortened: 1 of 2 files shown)


maleadt:
The base functionality (and corresponding tests) for this lives in GPUArrays.jl, https://github.com/JuliaGPU/GPUArrays.jl/blob/79cc61b1752523ce8386fe2904cc51a27e187c26/src/host/abstractarray.jl#L42-L57, so maybe we should start by updating it there so that we always show `CuArray(...)`? And/or adapt the test at https://github.com/JuliaGPU/GPUArrays.jl/blob/79cc61b1752523ce8386fe2904cc51a27e187c26/test/testsuite/base.jl#L239-L240 so that it only checks for the presence of `[1]` etc to make the `cu` optimization here possible.

PR #1215, opened 2021-10-22
improve sparse matrix conversions

Roger-luo:
this PR improves the sparse array conversion taking account of the indices type to support explicitly declared type conversion `CuSparseType{Tv, Ti}(dS)`

A few things I'm not sure how to deal with yet:

1. how do we convert `CSR{Tv, Int64}` to CPU types and CSC? it seems CUDA doesn't support this yet, and it seems costing a lot of extra computation to convert CSR to CSC on CPU just for printing? maybe we should have a method `cpu` that convert CSR/BSR to transpose(CSC) on CPU?

### DIFF
--- a/lib/cusparse/array.jl
+++ b/lib/cusparse/array.jl
@@ -26,6 +26,13 @@ mutable struct CuSparseVector{Tv, Ti} <: AbstractCuSparseVector{Tv, Ti}
     end
 end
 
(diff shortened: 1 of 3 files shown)


maleadt:
>     1. maybe we should have a method `cpu` that convert CSR/BSR to transpose(CSC) on CPU?

That seems like the cheapest solution for printing, yes.

PR #1273, opened 2021-12-10
Add const keyword to type aliases declarations.

eliascarv:
I noticed that the type aliases were not declared with the const keyword and I decided to do this PR to add them.

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -117,9 +117,9 @@ end
 
 ## convenience constructors
 
-CuVector{T} = CuArray{T,1}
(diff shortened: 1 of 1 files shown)


maleadt:
I thought that was only necessary back in the days?

eliascarv:
Yes, but declaring it as a constant prevents the identifier from being assigned another value.
e. g.:
image

eliascarv:
Also, non-constant global variables are not good for code performance.
e. g.:
image

maleadt:
The performance improvement alone is reason enough to do it like that. Thanks for the PR!

PR #1437, opened 2022-03-09
Initial wrap of cuquantum

kshyatt:
This is an initial wrap of the [cuQuantum SDK](https://developer.nvidia.com/cuquantum-sdk).

We'll need to do some artifact builds and then update the `deps` scripts. Not everything in `CUSTATEVEC` is wrapped, but you can take samples and compute expectation values!


maleadt:
At some point we'll have to move things in subpackages again; CUDA.jl is just becoming too massive to load. I wonder of cuQuantum, which presumably is pretty disconnected from any stdlib libraries that are loaded by default, would be a good candidate for that?

kshyatt:
Yeah, could be. We could probably just make it a subpackage in the same repo?

maleadt:
Yeah I definitely want to avoid fragmentation across repositories. Maybe Pkg is capable enough to handle multi-package repositories now? Worth exploring (but that also means we'll need to think about how to make some of the functionality, like the handle caches or APIUtils, work across packages).

kshyatt:
ITensors.jl and Makie seem to be handling it happily enough? Maybe we could
ask their thoughts?

kshyatt:
OK, the two modules are now split into their own subpackages! Tests passed for each running locally but I'm not quite sure how to deal with `bin/deps.jl` and friends as well as the CI scripts.

maleadt:
> I'm not quite sure how to deal with `bin/deps.jl` and friends as well as the CI scripts.

I guess we can still manage the binary dependencies centrally from the CUDA.jl module (they shouldn't impact load time much). What's the issue with the CI scripts?

kshyatt:
The CI script issue is mostly that I haven't dealt with any CI in ages 😓

.buildkite/pipeline.yml
```diff
@@ -300,6 +300,50 @@ steps:
...
+          codecov: true
+          dirs:
+            - src
+            - lib
+            - examples
+    agents:
+      queue: "juliagpu"
+      cuda: "11.0"
+      cap: "sm_80"
+    package: "CUSTATEVEC.jl"
```
maleadt: This should be a property of the julia-test plugin, not of the step itself.

kshyatt:
Did I manage to do the build kite wrong again?

.buildkite/pipeline.yml
```diff
@@ -197,6 +197,50 @@ steps:
...
+            - examples
+    agents:
+      queue: "juliagpu"
+      cuda: "11.0"
+      cap: "sm_80"
+    env:
+      JULIA_CUDA_VERSION: '11.6'
+      JULIA_CUDA_USE_BINARYBUILDER: 'true'
+    if: build.message !~ /\[skip tests\]/ && !build.pull_request.draft
+    timeout_in_minutes: 120
```
DilumAluthge:
```suggestion
  - label: "CUTENSORNET"
    plugins:
      - JuliaCI/julia#v1:
          version: 1.6
      - JuliaCI/julia-coverage#v1:
          codecov: true
          dirs:
            - src
            - lib
            - examples
    agents:
      queue: "juliagpu"
      cuda: "11.0"
      cap: "sm_80"
    env:
      JULIA_CUDA_VERSION: '11.6'
      JULIA_CUDA_USE_BINARYBUILDER: 'true'
    if: build.message !~ /\[skip tests\]/ && !build.pull_request.draft
    timeout_in_minutes: 120
    commands: |
      export JULIA_PROJECT="CUTENSORNET"
      export JULIA_LOAD_PATH="@"
      julia -e 'import Pkg; Pkg.instantiate()'
      julia -e 'import Pkg; Pkg.precompile()'
      julia -e 'import Pkg; Pkg.test()'
```

.buildkite/pipeline.yml
```diff
@@ -197,6 +197,54 @@ steps:
...
+    commands: |
+      export JULIA_PROJECT="CUTENSORNET"
+      export JULIA_LOAD_PATH="@"
```
DilumAluthge:
```suggestion
```

This isn't actually necessary.

When we do `Pkg.test`, Pkg will run the tests in a subprocess that has `JULIA_LOAD_PATH=@`.

DilumAluthge:
@kshyatt Of the subdir packages in this repository, which packages depend on other packages? I'll need to `Pkg.dev` the dependencies before building the package itself, since we want to make sure to use the dev version of all dependencies.

kshyatt:
each of `CUTENSORNET` and `CUSTATEVEC` depend only on `CUDA`

.buildkite/pipeline.yml
```diff
@@ -241,6 +239,12 @@ steps:
       JULIA_CUDA_USE_BINARYBUILDER: 'true'
     if: build.message !~ /\[skip tests\]/ && !build.pull_request.draft
     timeout_in_minutes: 120
+    commands: |
+      export JULIA_PROJECT="CUTENSORNET"
```
maleadt: Is this equivalent to the `--project` argument? If so, can't we use that option for the buildkite plugin? I'd prefer that over duplicating the commands here.

.buildkite/pipeline.yml
```diff
@@ -197,6 +197,56 @@ steps:
...
+    env:
+      JULIA_CUDA_VERSION: '11.6'
+      JULIA_CUDA_USE_BINARYBUILDER: 'true'
+    if: build.message !~ /\[skip tests\]/ && !build.pull_request.draft
+    timeout_in_minutes: 120
+    commands: |
+      export JULIA_PROJECT="CUTENSORNET"
+      julia -e 'import Pkg; Pkg.instantiate()'
+      julia -e 'import Pkg; Pkg.precompile()'
+      julia -e 'import Pkg; Pkg.test()'
```
DilumAluthge:
```suggestion
  - label: "CUSTATEVEC"
    plugins:
      - JuliaCI/julia#v1:
          version: 1.6
      - JuliaCI/julia-coverage#v1:
          codecov: true
          dirs:
            - src
            - lib
            - examples
    agents:
      queue: "juliagpu"
      cuda: "11.0"
      cap: "sm_80"
    env:
      JULIA_CUDA_VERSION: '11.6'
      JULIA_CUDA_USE_BINARYBUILDER: 'true'
      PACKAGE: 'CUSTATEVEC'
    if: build.message !~ /\[skip tests\]/ && !build.pull_request.draft
    timeout_in_minutes: 120
    commands: |
      julia -e 'import Pkg; Pkg.develop(; path = pwd())' # CUDA
      julia -e 'import Pkg; Pkg.develop(; path = joinpath(pwd(), ENV["PACKAGE"]))' # CUSTATEVEC
      julia -e 'import Pkg; Pkg.precompile()'
      julia -e 'import Pkg; Pkg.test(ENV["PACKAGE"])'

  - label: "CUTENSORNET"
    plugins:
      - JuliaCI/julia#v1:
          version: 1.6
      - JuliaCI/julia-coverage#v1:
          codecov: true
          dirs:
            - src
            - lib
            - examples
    agents:
      queue: "juliagpu"
      cuda: "11.0"
      cap: "sm_80"
    env:
      JULIA_CUDA_VERSION: '11.6'
      JULIA_CUDA_USE_BINARYBUILDER: 'true'
      PACKAGE: 'CUTENSORNET'
    if: build.message !~ /\[skip tests\]/ && !build.pull_request.draft
    timeout_in_minutes: 120
    commands: |
      julia -e 'import Pkg; Pkg.develop(; path = pwd())' # CUDA
      julia -e 'import Pkg; Pkg.develop(; path = joinpath(pwd(), ENV["PACKAGE"]))' # CUSTATEVEC
      julia -e 'import Pkg; Pkg.precompile()'
      julia -e 'import Pkg; Pkg.test(ENV["PACKAGE"])'
```

.buildkite/pipeline.yml
DilumAluthge: Yeah, `JULIA_PROJECT` and `--project` are basically equivalent.

But, I just made a new suggestion that doesn't use `JULIA_PROJECT` or `--project`. Because we need to `dev` both `CUDA` and `CUTENSORNET` (resp. `CUSTATEVEC`) in the same environment.

maleadt: Ah yes, of course. That's always as bit of a mess, e.g., how we do reverse-CI in GPUCompiler.jl: https://github.com/JuliaGPU/GPUCompiler.jl/blob/b73a0c24bfd01fb58a57048711dfb341a56c7d31/.buildkite/pipeline.yml#L2-L22=

DilumAluthge: Yeah. We could try to add a feature to the Buildkite plugin that would automatically `Pkg.dev` the dependencies, but this is a relatively uncommon use case, so I don't know if it's worth doing.

kshyatt:
lol i forgot to add `Test`, let me fix that

DilumAluthge:
Looks like it's not actually able to find `CUSTATEVEC` (resp. `CUTENSORNET`). Is this a problem with our test setup?

DilumAluthge:
@maleadt What do we do if we want to print the verbose output while searching for libraries?

Would `export JULIA_DEBUG="CUDA,CUSTATEVEC,CUTENSORNET"` be sufficient to get all of the verbose output?

kshyatt:
It needs the artifacts,probably

DilumAluthge:
Looks like you added the artifacts to `Artifacts.toml` already though, right? And they're not `lazy`, so they should have been downloaded.

maleadt:
Artifacts are handled by CUDA.jl, so `JULIA_DEBUG=CUDA` should suffice. The artifacts should be lazy though so that they only get downloaded when the libraries are used.

kshyatt:
So is the issue likely to be me writing the `deps/` code for these artifacts wrong?

maleadt:
Did you test it locally with artifacts (i.e., it is a CI-only issue)?

kshyatt:
Might be a case mismatch hehe

kshyatt:
Tests passed and we can register these two later. Squashing and merging.

maleadt:
This should have been squashed...

PR #1467, opened 2022-04-07
CUSPARSE: Better error msg for unsupported sparse mm

Roger-luo:
currently when the element-type happens to be mix of `Float32` and `Float64` (which happens quite often for users when they use `cu` for the conversion) `mul!` will just call the generic fallback and if `CUDA.allowscalar` is not set, this would just throw a warning a while later.

This PR let this dispatch path just error since it's not supported for now anyways.

### DIFF
--- a/lib/cusparse/interfaces.jl
+++ b/lib/cusparse/interfaces.jl
@@ -54,6 +54,35 @@ for (taga, untaga) in tag_wrappers, (wrapa, transa, unwrapa) in op_wrappers
     end
 end
 
+# generic fallback error
+tag_wrappers = ((identity, identity),
+                (T -> :(HermOrSym{<:Any, <:$T}), A -> :(parent($A))))
+op_wrappers = (
+    (identity, T -> 'N', identity),
+    (T -> :(Transpose{<:Any, <:$T}), T -> 'T', A -> :(parent($A))),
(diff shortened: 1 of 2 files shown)


Roger-luo:
it seems CI failure unrelated to this PR

maleadt:
Although I like better error messages, which is why I filed https://github.com/JuliaGPU/CUDA.jl/issues/1271, I don't think that adding a lengthy implementation like this for every operation scales well. We need a simpler way to add such implementations (metaprogramming) or another way to detect the use of fallbacks. Isn't the scalar indexing error message clear enough?

Roger-luo:
> Isn't the scalar indexing error message clear enough?

I think scalar indexing is just a general error for CUDA since it doesn't say anything about why it is scalar indexing, e.g most of cases I'd like to know whether it's because something is not implemented. 

Yeah, I'm not sure what would be the best way to detect this type of interface overloading automatically, so I just did it manually since I hit this one quite often...

Roger-luo:
OK just had another idea, what if we make a macro `@not_supported` for this? e.g

```julia
@not_supported <method signature>
```

this is probably more readable?

maleadt:
Yeah that seems like a better approach.

PR #1520, opened 2022-05-22
RFC: Support Adapt.AbstractGPUDevice

oschulz:
See JuliaGPU/Adapt.jl#52

Functional (not tests yet).


oschulz:
Closed, but not forgotten, to be reincarnated in the future (see JuliaGPU/Adapt.jl#64).

PR #1544, opened 2022-06-14
Explicit tests for type conversion

kshyatt:
A bunch of these paths weren't covered. Just wanted to make sure we don't have or introduce any typos.


maleadt (approved):
LGTM, thanks!

PR #1612, opened 2022-10-05
[CUSPARSE] Use cusparseSpMM_preprocess

amontoison:
The function `cusparseSpMM_preprocess()` can be called before cusparseSpMM to speedup the actual computation. It is useful when cusparseSpMM is called multiple times with the same sparsity pattern (matA). The values of the matrices (matA, matB, matC) can change arbitrarily. It provides performance advantages if used with `CUSPARSE_SPMM_CSR_ALG1` or `CUSPARSE_SPMM_CSR_ALG3`. For all other formats and algorithms have no effect. 

https://docs.nvidia.com/cuda/cusparse/index.html#cusparse-generic-function-spmm


amontoison:
It's not working for CUDA v"11.5"...

PR #1603, opened 2022-09-24
Added more sparse functions like: kron, tril, triu, reshape, adjoint, transpose, sparse-sparse multiplication

albertomercurio:
I added more functions to support the sparse CUDA arrays.

What I added:

- `kron` function, which returns the kronecker product of two sparse matrices.
- `tril` and `triu` which return the lower-tringular and upper-triangular of a sparse matrix. I is possible also to insert an offset `k`, as usual.
- sparse-sparse multiplication, which was missing before, and I think is the most important feature I added.
- `adjoint` and `transpose` functions.
- `reshape` of sparse matrices.
- `opnorm` now supports also the `p=1` case.
- `norm` function with all cases (p=0,1,2,..,Inf,-Inf)
- `SparseArrays.droptol!` which drops all the elements smaller than a certain quantity
- `exp` function for matrix exponential. Often very useful for quantum mechanics calculations. It does not use the Pade approximation for several reasons. However this method seems to be very stable and fast.
- `ishermitian` and `issymmetric` support. Before it was returning always false for a generic CuSparseMatrix.

About dense matrices, I also added the support for the LinearAlgebra `eigen` function.

PR #1677, opened 2022-11-21
Fix eigen when using Hermitian or Symmetric matrices

GVigne:
Hi!
I tried to use the implementation of `eigen` from #1603 but ran into some issues.
First, trying to use `eigen` with a `Hermitian` or `Symmetrix` matrix fails with the following error:
`MethodError: no method matching syevd!(::Char, ::Char, ::Symmetric{Float32, CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}}).`
This is because calling `copy(A)` if `A` is `Hermitian` or `Symmetric` will also copy the wrapper around it, resulting in the creation of a new `Hermitian` or `Symmetric` array and not a plain array.
Second, those wrappers don't actually check if the given matrix really is hermitian/symmetric, so that running the following code will not bug:
```
julia> A = rand(10,10); # A is a matrix of Float64, and probably not symmetric
julia> B = Hermitian(A);
```
However, CUDA's `heevd` and `syevd` routines are specialised (for complex and real numbers respectively), so we can't simply call `heevd` if we are dealing with a `Hermitian` matrix, and `syevd` if we are dealing with `Symmetric` matrices.

This PR aims to fix this by copying the array hidden behind those structures, and by creating a fallback if `eigen` has to deal with a real matrix wrapped in `Hermitian`. 

I also added a few tests.

### DIFF
--- a/lib/cusolver/linalg.jl
+++ b/lib/cusolver/linalg.jl
@@ -72,13 +72,17 @@ end
 # eigenvalues
 
 function LinearAlgebra.eigen(A::Symmetric{T,<:CuMatrix}) where {T<:BlasReal}
-    A2 = copy(A)
+    A2 = copy(A.data)
     Eigen(syevd!('V', 'U', A2)...)
 end
-function LinearAlgebra.eigen(A::Hermitian{T,<:CuMatrix}) where {T<:BlasFloat}
-    A2 = copy(A)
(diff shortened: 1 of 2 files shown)


maleadt (approved):
LGTM, thanks!

PR #1809, opened 2023-03-16
Initialize globals to nothing

simonbyrne:
Alternative fix to #1798; requires https://github.com/JuliaPackaging/Yggdrasil/pull/6398

### DIFF
--- a/src/initialization.jl
+++ b/src/initialization.jl
@@ -36,9 +36,9 @@ function __init__()
     #       and remove functional(), once people sufficiently use weak dependencies.
 
     # check that we have a driver
-    global libcuda
+    global libcuda = nothing
     if CUDA_Driver_jll.is_available()
-        if isdefined(CUDA_Driver_jll, :libcuda)
+        if isdefined(CUDA_Driver_jll, :libcuda) && !isnothing(CUDA_Driver_jll.libcuda)
             libcuda = CUDA_Driver_jll.libcuda
         else
             _initialization_error[] = "CUDA driver not found"
--- a/src/initialization.jl
+++ b/src/initialization.jl
@@ -38,7 +38,7 @@ function __init__()
     # check that we have a driver
     global libcuda = nothing
     if CUDA_Driver_jll.is_available()
-        if isdefined(CUDA_Driver_jll, :libcuda) && !isnothing(CUDA_Driver_jll.libcuda)
+        if !isnothing(CUDA_Driver_jll.libcuda)
             libcuda = CUDA_Driver_jll.libcuda
         else
             _initialization_error[] = "CUDA driver not found"
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -34,9 +34,9 @@ version = "0.4.2"
 
 [[CUDA_Driver_jll]]
 deps = ["Artifacts", "JLLWrappers", "LazyArtifacts", "Libdl", "Pkg"]
-git-tree-sha1 = "8b912e69a6826d32eced5c4f9e2c50aff7100979"
+git-tree-sha1 = "ccf9b16de0c251cb3e5965b6e0fcc224e0acdf36"
 uuid = "4ee394cb-3365-5eb0-8335-949819d2adfc"
-version = "0.4.0+0"
+version = "0.4.0+1"
 
 [[CUDA_Runtime_Discovery]]
 deps = ["Libdl"]


src/initialization.jl
```diff
@@ -36,9 +36,9 @@ function __init__()
     #       and remove functional(), once people sufficiently use weak dependencies.
 
     # check that we have a driver
-    global libcuda
+    global libcuda = nothing
     if CUDA_Driver_jll.is_available()
-        if isdefined(CUDA_Driver_jll, :libcuda)
+        if isdefined(CUDA_Driver_jll, :libcuda) && !isnothing(CUDA_Driver_jll.libcuda)
```
maleadt: Why keep the `isdefined`? CUDA_Driver_jll@0.4 isn't used by anybody yet, so we can just keep the second path

maleadt:
Aren't all JLLs susceptible to this? Or only ones that set things from `__init__`?

vchuravy:
Only things that are set from _init_

maleadt:
Then we should probably also rework CUDA_Runtime_Discovery_jll, which does similar `__init__`-time shenanigans

src/initialization.jl
simonbyrne: I wasn't sure. That's good to know!

simonbyrne:
```suggestion
        if !isnothing(CUDA_Driver_jll.libcuda)
```

maleadt:
```
   Installed CUDA_Driver_jll ─ v0.4.0+1
ERROR: LoadError: MethodError: no method matching dlopen(::Nothing; throw_error=false)
Closest candidates are:
  dlopen(!Matched::AbstractString) at libdl.jl:116 got unsupported keyword argument "throw_error"
  dlopen(!Matched::AbstractString, !Matched::Integer; throw_error) at libdl.jl:116
  dlopen(!Matched::Function, !Matched::Any...; kwargs...) at libdl.jl:141
  ...
Stacktrace:
 [1] cuda_toolkit_tag()
   @ Main ~/.cache/julia-buildkite-plugin/depots/3cc01fab-3357-4a7a-9294-cde2d3115a97/packages/CUDA_Runtime_jll/cq9MZ/.pkg/platform_augmentation.jl:94
 [2] augment_platform!(platform::Platform)
   @ Main ~/.cache/julia-buildkite-plugin/depots/3cc01fab-3357-4a7a-9294-cde2d3115a97/packages/CUDA_Runtime_jll/cq9MZ/.pkg/platform_augmentation.jl:137
 [3] top-level scope
   @ ~/.cache/julia-buildkite-plugin/depots/3cc01fab-3357-4a7a-9294-cde2d3115a97/packages/CUDA_Runtime_jll/cq9MZ/.pkg/select_artifacts.jl:11
 [4] include(fname::String)
   @ Base.MainInclude ./client.jl:476
 [5] top-level scope
   @ none:5
in expression starting at /root/.cache/julia-buildkite-plugin/depots/3cc01fab-3357-4a7a-9294-cde2d3115a97/packages/CUDA_Runtime_jll/cq9MZ/.pkg/select_artifacts.jl:11
```

maleadt:
Had to revert the CUDA_Driver_jll change because it somehow did enter user environments, and breaks CUDA_Runtime_jll as shown above. I guess we'll have to be defensive and fix-up all uses first before making this change / make it a breaking change.

simonbyrne:
Sorry about that.

maleadt:
No problem, my understanding of the impact of the change was wrong too.

maleadt:
The culprit turned out to be XGBoost_jll, which pulls in CUDA_Runtime_jll. However, without a dependency on CUDA.jl, any version of these JLLs could be installed, i.e., there was no CUDA.jl constraining it to CUDA_Driver_jll 0.2.

This is an interesting case though. It means that users who aren't interested in CUDA.jl might still end up with CUDA_Runtime_jll loading CUDA libraries because of some JLL having GPU capabilities. I guess that's not an actual problem though, because I would assume that the GPU-capable version of XGBoost_jll still exposes the CPU functionality the user is interested in.

vchuravy:
> This is an interesting case though. I

Similarly for PAPI. This is why I added `cuda=none`, but if you install PAPI_jll@7 on a machine with a CUDA GPU we will pull in the newer versions

maleadt:
OK, AFAIU the path forward here is to:
- update CUDA_Driver_jll like in https://github.com/JuliaPackaging/Yggdrasil/pull/6398, but this time with a breaking version bump, 0.5.
- update CUDA_Runtime_jll's augmentation to be compatible with this (i.e., compat bounds, and fixing https://github.com/JuliaPackaging/Yggdrasil/blob/7e6c6ca90991b448c7fb32f205c665270d982039/C/CUDA/CUDA_Runtime/platform_augmentation.jl#L50), and tag 0.5 (also a breaking release)
- optionally, fix CUDA_Runtime_Discovery_jll
- update this PR

maleadt:
This was fixed in https://github.com/JuliaGPU/CUDA.jl/pull/1832.

PR #1826, opened 2023-03-23
Add Julian API for NVPerf and CUPTI Profiler

vchuravy:


maleadt:
In terms of a high-level API, it'd be super nice to have something a la `Base.@timed`, returning requested performance counters:

```julia
julia> using NVPerf
julia> NVPerf.@measured dram__bytes.sum.per_second begin
         @cuda kernel()
       end
(dram_bytes = (sum = (per_second = 1.0f0,),),)
```

PR #2159, opened 2023-11-09
[CUSPARSE] Update the wrapper of cusparseSpSV_updateMatrix

amontoison:
I also added two constructors.
They are needed if we want to perform the analysis of sparse triangular solves (`spsv` / `spsm`) without providing the right-hand side(s).


maleadt (approved):
LGTM assuming these will get test coverage later.

PR #2215, opened 2024-01-03
Add a note explaining that Pkg.test can cause compilation errors and how to fix.

kshyatt:
Thank you @giordano for the hint. Thought it would be nice to leave a note about this for others.

### DIFF
--- a/docs/src/development/troubleshooting.md
+++ b/docs/src/development/troubleshooting.md
@@ -38,6 +38,13 @@ e.g., due to inaccurate or instable type information). These are a red herring, 
 cause is listed last: a typo in the use of the `threadIdx` function! If we fix this, the IR
 error disappears and our kernel successfully compiles and executes.
 
+!!! note 
+
+    Bounds checking can cause compilation failures. Even if you have included an `@inbounds` block,
+    `Pkg.test` by default will force bounds-checking and this may cause `InvalidIRError`s. To prevent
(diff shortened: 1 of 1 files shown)


maleadt:
Do you have any concrete MWE? Bounds errors, and by extension all errors, are supposed to be GPU compatible because of the quirks we have (which get rid of allocations and/or replace the exception with a GPU compatible one): https://github.com/JuliaGPU/CUDA.jl/blob/master/src/device/quirks.jl. If some error isn't, that probably warrants a new quirk.

kshyatt:
Posted about it on slack, I’ll DM you

maleadt:
As discussed on Slack: I don't like this advice. Our goal is to support as much of the language as possible, and it shouldn't be required to use `@inbounds` (and conversely test with `--check-bounds=auto` or `--check-bounds=no`) for code to compile for the GPU. I'd rather users file issues about that, so that we can fix those errors by adding adequate quirks.

Of course, if that turns out impossible for some reason or another, it would be fine to add something like this to the documentation. But my understanding right now that we can make exceptions GPU compatible using overlay methods, so for now I propose we don't add this to the docs.

That said, I don't feel too strongly about this, so if anybody *really* wants this in the docs, feel free to comment or re-open.

maleadt:
Also, @kshyatt, given that your package seems to require this (for now at least) you can do something way more hackier: https://github.com/JuliaGPU/CUDAnative.jl/commit/f7d7a984d4aa23894a80f69dddb1cf6a19f3d147

PR #2264, opened 2024-02-13
cuTENSOR: refactor obtaining compute_type as part of plan

lkdvos:
Removes the duplicate code in obtaining the `compute_type` which was implemented both in `execute_*operation*` and `plan_*operation*`. Now the keyword argument is just passed through to `plan_*operation*` and determined there.

This implements the refactor discussed in #2263.


maleadt (approved):
Thanks!

PR #2270, opened 2024-02-16
[CUBLAS] Use the ILP64 API with CUDA 12

### LINKED ISSUE #2268: CUBLAS: nrm2 support for StridedCuArray with length requiring Int64

amontoison:
close #2268


tipfom:
Thanks! Very nice 🥳

PR #2368, opened 2024-05-08
Mark cublas version/handle as non-differentiable

wsmoses:
I'm working on getting a minimal working cublas AD working, but these are nevertheless prerequisites that will reduce the error log size for a lot of users already (e.g. https://github.com/EnzymeAD/Enzyme.jl/issues/1392)

### DIFF
--- a/ext/EnzymeCoreExt.jl
+++ b/ext/EnzymeCoreExt.jl
@@ -13,6 +13,14 @@ else
     using ..EnzymeCore.EnzymeRules
 end
 
+function EnzymeCore.EnzymeRules.inactive(::typeof(CUDA.CUBLAS.handle))
+    return nothing
+end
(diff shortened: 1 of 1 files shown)


vchuravy (approved):
Why does cudaconvert need a forward pass?

wsmoses:
It needed one for the other PR to ensure unified memory setup is fine

wsmoses:
@vchuravy can you merge?

vchuravy:
> can you merge?

Yes, but there is a question how we test this code.

wsmoses:
Yeah I have a test [just doing CUBLAS.Dot(x, y) but that presently fails for the reason of cudamemset as a fn not being found so we need to figure out the correct symbol materialization story. Still helpful for reducing error before then tho

PR #2458, opened 2024-08-05
Implementation of spdiagm for CUSPARSE

### LINKED ISSUE #1857: Creating a sparse diagonal matrix of CuArray(u)
If we have a vector (u), we can easily convert it into a sparse diagonal matrix using the command spdiagm(u).

If we want to do the same operation on the GPU using CUDA where u is defined as CuArray(u). The command spdiagm(u) gives the scalar indexing error and therefore it does not work.

 CuSparseMatrixCSC(Diagonal(u)) works but it is quite slow. Is there a way to perform this task efficiently?

Regards

walexaindre:
Here is the implementation of SparseArrays.spdiagm(vec)

The one(Int32) and Int32 conversions are used to reduce allocations and enhance the performance of this simple function.

Closes #1857

### DIFF
--- a/lib/cusparse/array.jl
+++ b/lib/cusparse/array.jl
@@ -324,7 +324,7 @@ end
 
 function SparseArrays.spdiagm(v::CuArray{Tv}) where {Tv}
     nzVal = v
-    N = Int32(length(v))
+    N = Int32(length(nzVal))
     
     colPtr = CuArray(one(Int32):(N + one(Int32)))
     rowVal = CuArray(one(Int32):N)
--- a/lib/cusparse/array.jl
+++ b/lib/cusparse/array.jl
@@ -322,7 +322,7 @@ function SparseArrays.sparsevec(I::CuArray{Ti}, V::CuArray{Tv}, n::Integer) wher
     CuSparseVector(I, V, n)
 end
 
-function SparseArrays.spdiagm(v::CuArray{Tv}) where {Tv}
+function SparseArrays.spdiagm(v::CuVector{Tv}) where {Tv}
     nzVal = v
     N = Int32(length(nzVal))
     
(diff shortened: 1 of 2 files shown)


lib/cusparse/array.jl
```diff
@@ -322,6 +322,16 @@ function SparseArrays.sparsevec(I::CuArray{Ti}, V::CuArray{Tv}, n::Integer) wher
     CuSparseVector(I, V, n)
 end
 
+function SparseArrays.spdiagm(v::CuArray{Tv}) where {Tv}
```
amontoison:
```suggestion
function SparseArrays.spdiagm(v::CuVector{Tv}) where {Tv}
```

amontoison (changes requested):
Can you add a test for your implementation of `spdiagm`?

walexaindre:
Thank you @amontoison for looking at my PR. I can implement that test but I am confused about where I need to do that exactly. 

Can you suggest me a possible file where I can write that test?

I was thinking about generic.jl or interfaces.jl. However, generic.jl looks like a place where some basic operations between matrix and vectors are tested and interfaces.jl looks like a file where linear algebra related functions are the main theme. 

I belive that interfaces.jl could be the correct place. But if you can shed some light on the logic behind file organization and where it is expected to implement something like this, I would appreciate it.

walexaindre:
For now, tests are implemented as a new test section in interfaces.jl

walexaindre (commented):
Fixed

amontoison (approved):

amontoison:
> For now, tests are implemented as a new test section in interfaces.jl

The test files `interfaces.jl` and `linalg.jl` are fine.
LGTM!

PR #2538, opened 2024-10-30
Removed allocations for transpose/adjoint - diagonal multiplications

mkhrenov:
The dgmm! based implementation of multiplication of transpose/adjoint matrices with diagonal matrices that got merged makes a call to CuMatrix on the transpose/adjoint. This results in allocations and a substantial performance penalty. I haven't been able to figure out how to do this in place using the cuBLAS functions, but writing it out as in-place broadcast operations seems to work well. Open to any suggestions.


lib/cublas/linalg.jl
```diff
@@ -347,19 +347,27 @@ function LinearAlgebra.mul!(C::CuMatrix{T}, A::Diagonal{T,<:CuVector}, B::CuMatr
 end
 
 function LinearAlgebra.mul!(C::CuMatrix{T}, A::Transpose{T,<:CuMatrix}, B::Diagonal{T,<:CuVector}) where {T<:CublasFloat}
-    return dgmm!('R', CuMatrix(A), B.diag, C)
+    C .= A
+    C .*= B.diag'
```
amontoison:
```suggestion
    C .*= transpose(B.diag)
```

amontoison: You don't want to conjugate the coefficients if `B` is a complex diagonal matrix.

lib/cublas/linalg.jl
```diff
@@ -347,19 +347,27 @@ function LinearAlgebra.mul!(C::CuMatrix{T}, A::Diagonal{T,<:CuVector}, B::CuMatr
...
-    return dgmm!('L', CuMatrix(B), A.diag, C)
+    C .= B
+    C .*= A.diag
+    return C
 end
 
 function LinearAlgebra.mul!(C::CuMatrix{T}, A::Adjoint{T,<:CuMatrix}, B::Diagonal{T,<:CuVector}) where {T<:CublasFloat}
-    return dgmm!('R', CuMatrix(A), B.diag, C)
+    C .= A
+    C .*= B.diag'
```
amontoison: Why do you conjugate `B`?

amontoison (changes requested):
I don't understand how you do some products, `M * D` and `D * M` gives different results and based of your broadcast relations, you can only do one them.
You use the trick of the adjoint / transpose like `D * M = (M' * D')'` but it means that you need to conjugate the result `C` at the end.

If you find a way to do that such that all tests passed, it's fine for me to use broadcast.

amontoison:
I don't think that we can use `dgmm` without allocation for transpose/adjoint - diagonal multiplications because we can't reinterpret the input matrices as matrices stored in row-major format. So broadcast is the only alternative if we don't want allocations.

lib/cublas/linalg.jl
amontoison: Hum, I think you use a trick that I don't understand because you don't transpose `A`.
Can you explain what you do exactly?

mkhrenov: Sure, so the transpose (which I've put in both functions) of B.diag isn't a result of A being Transpose or Adjoint, it's just to make the broadcast multiplication work on the correct dimension for B diagonal. A being Adjoint or Transpose gets realized during the broadcasted assignment to C in all cases.

lib/cublas/linalg.jl
mkhrenov: It should have been a transpose, fixed.

amontoison (approved):

mkhrenov:
Benchmarks seem to have been interrupted for some reason, but everything else looks good. Anything else I should do?

amontoison:
LGTM but I can't merge the PR if all tests didn't passed.
You need Tim to manually merge the PR.

PR #2541, opened 2024-11-06
Consistent use of Nsight Compute

huiyuxie:
No big issue but I prefer it to be consistent.


huiyuxie:
Please help me add docs label

PR #2565, opened 2024-11-26
Update EnzymeCoreExt.jl

simenhu:
The inactive_noinl failed in broadcast assignment when using Enzyme=0.13.16 and CUDA=5.5.2. Switching to the version that prevents inlining seem to have fixed it. Are there any negative consequences of this?

The issue where the error can be reproduced can be found here: EnzymeAD/Enzyme.jl#2116

### DIFF
--- a/ext/EnzymeCoreExt.jl
+++ b/ext/EnzymeCoreExt.jl
@@ -25,7 +25,7 @@ end
 function EnzymeCore.EnzymeRules.inactive_noinl(::typeof(CUDA.is_pinned), args...)
     return nothing
 end
-function EnzymeCore.EnzymeRules.inactive_noinl(::typeof(CUDA.launch_configuration), args...; kwargs...)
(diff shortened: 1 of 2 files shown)


wsmoses:
Can you add your full code as a test?

ext/EnzymeCoreExt.jl
ptiede, line 1: This code should test the change

```julia
@testset "Reverse sum abs2" begin
    x = CuArray([1.0, 2.0, 3.0, 4.0])
    dx = CuArray([0., 0.0, 0.0, 0.0])
    f(x) = sum(abs2.(x))
    Enzyme.autodiff(Reverse, f, Active, Duplicated(x, dx))
    @test all(dx .≈ 2.*x)
end
```

simenhu:
> Can you add your full code as a test?

Do you still want me to add the code as an test? Assume that you want me to remove the dependence on OrdinaryDiffEq?

wsmoses:
yeah I don't think we should add a dependency here, but if you can make a cuda exclusive test, that would be good to keep

PR #2574, opened 2024-12-02
[CUSOLVER] Add more tests for the dense SVD

amontoison:
Update: I need to check the size of the buffers.
The features not tested are not working.

### DIFF
--- a/lib/cusolver/dense.jl
+++ b/lib/cusolver/dense.jl
@@ -394,9 +394,9 @@ for (bname, fname, elty, relty) in ((:cusolverDnSgesvd_bufferSize, :cusolverDnSg
 
             U = if jobu === 'A'
                 similar(A, $elty, (m, m))
-            elseif jobu == 'S' || jobu === 'O'
+            elseif jobu === 'S'
                 similar(A, $elty, (m, k))
-            elseif jobu === 'N'
+            elseif jobu === 'N' || jobu === 'O'
                 CU_NULL
             else
                 error("jobu must be one of 'A', 'S', 'O', or 'N'")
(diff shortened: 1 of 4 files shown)


amontoison:
@maleadt The tests are failing because of the new release 1.11.2.
I suspect that the culprit is this PR in LinearAlgebra.jl: https://github.com/JuliaLang/julia/pull/56149

amontoison:
@jishnub and @dkarrasch, do you have a idea of how to fix it?
I expect that we will have the same issue in `AMDGPU.jl`.

dkarrasch:
The issue is caused by https://github.com/JuliaLang/LinearAlgebra.jl/commit/9d398baed01a39a26279d4564c254ce333458520. I think it could be fixed by adding `[l/r]mul!` methods (over at LinearAlgebra.jl) with `UpperOrLowerTriangular` signature, to bypass the `istriu` check, which comes with a price for `LowerTriangular`s, additionally to the scalar indexing.

amontoison:
@pxl-th fixed the issue in `AMDGPU.jl` by adding the following lines:
```julia
const ROCUpperOrUnitUpperTriangular = LinearAlgebra.UpperOrUnitUpperTriangular{
    <:Any,<:Union{<:ROCArray, Adjoint{<:Any, <:ROCArray}, Transpose{<:Any, <:ROCArray}}}
const ROCLowerOrUnitLowerTriangular = LinearAlgebra.LowerOrUnitLowerTriangular{
    <:Any,<:Union{<:ROCArray, Adjoint{<:Any, <:ROCArray}, Transpose{<:Any, <:ROCArray}}}

LinearAlgebra.istriu(::ROCUpperOrUnitUpperTriangular) = true
LinearAlgebra.istril(::ROCUpperOrUnitUpperTriangular) = false
LinearAlgebra.istriu(::ROCLowerOrUnitLowerTriangular) = false
LinearAlgebra.istril(::ROCLowerOrUnitLowerTriangular) = true
```
@dkarrasch, do you recommend this solution for the GPU packages?

dkarrasch:
You can do that to quickly fix it for v1.11.2 (though it's very ugly), otherwise there's an open PR that fixes it. It will be backported and then released as v1.11.3.

PR #2577, opened 2024-12-06
[CUSOLVER] Interface XsyevBatched

amontoison:
Requested by @mtanneau.

I found a bug with `uplo = 'L'` and the precision `ComplexF32`, I should open a ticket.
I commented the related tests.

### DIFF
--- a/test/libraries/cusolver/dense_generic.jl
+++ b/test/libraries/cusolver/dense_generic.jl
@@ -225,4 +225,32 @@ p = 5
             d_W, neig = CUSOLVER.Xsyevdx!('N', 'V', uplo, d_A, vl=3.5, vu= 7.5, il=1, iu=3)
         end
     end
+
+    @testset "syevBatched!" begin
+        batch_size = 5
+        for uplo in ('L', 'U')
+            A = rand(elty, n, n * batch_size)
+            B = rand(elty, n, n * batch_size)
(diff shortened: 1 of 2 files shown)


test/libraries/cusolver/dense_generic.jl
```diff
@@ -215,4 +215,34 @@ p = 5
             d_W, neig = CUSOLVER.Xsyevdx!('N', 'V', uplo, d_A, vl=3.5, vu= 7.5, il=1, iu=3)
         end
     end
+
+    @testset "syevBatched!" begin
+        batch_size = 5
+        for uplo in ('L', 'U')
+            (uplo == 'L') && (elty == ComplexF32) && continue
```
mtanneau: Should this have a `@test_broken` or something similar to avoid forgetting about it in the future?

mtanneau (commented):
Thank you @amontoison! 🙏

I need to learn how to do this myself 🙃

test/libraries/cusolver/dense_generic.jl
amontoison: The issue is that one EVD in the batch randomly fails.
Maybe I should return an error directly in the function if someone wants to use it with this combinaison of parameters (`uplo = 'L'` and `T = ComplexF32`).

PR #2616, opened 2025-01-10
Switch CUBLAS to device-side pointer mode

kshyatt:
Attempting to address https://github.com/JuliaGPU/CUDA.jl/issues/2571

I've set the pointer mode to "device side" during handle creation. Since `gemmGroupedBatched` [doesn't support](https://docs.nvidia.com/cuda/cublas/#cublas-t-gemmgroupedbatched) device side pointer mode, it won't be usable. One workaround for this would be to add a new function to create a handle with host side mode, or add the pointer mode as an optional kwarg to `handle()`. Very open to feedback on this.

I've set this up so that users can supply `CuRef`s of the appropriate result type to the level 1 functions for results. If that's not provided, the functions execute as they do today (synchronously). Similarly, for functions taking `alpha` or `beta` scalar arguments, if the user provides `CuRef` (actually a `CuRefArray`), the functions will execute asynchronously and return instantly. If the user provides a `Number`, the behaviour is unchanged from today. I'm not married to this design and it can certainly be changed.

cc @Jutho

### DIFF
--- a/lib/cublas/wrappers.jl
+++ b/lib/cublas/wrappers.jl
@@ -1445,12 +1445,8 @@ for (fname, fname_64, elty) in ((:cublasSgemmGroupedBatched, :cublasSgemmGrouped
             Cptrs = unsafe_batch(reduce(vcat, C))
 
             try
-                ## XXX: does not seem to support device pointers
+                ## XXX: cublasXgemmGroupedBatched does not seem to support device pointers
                 cublasSetPointerMode_v2(handle(), CUBLAS_POINTER_MODE_HOST)
-
-                mode = Ref{cublasPointerMode_t}()
-                cublasGetPointerMode_v2(handle(), mode)
-                @show mode[]
 
                 if CUBLAS.version() >= v"12.0"
                     $fname_64(handle(), transa, transb, m, n, k, alpha, Aptrs, lda,
@@ -1507,12 +1503,19 @@ for (fname, fname_64, elty) in ((:cublasSgemmGroupedBatched, :cublasSgemmGrouped
             Bptrs = unsafe_batch(B)
             Cptrs = unsafe_batch(C)
 
-            if CUBLAS.version() >= v"12.0"
-                $fname_64(handle(), transa, transb, m, n, k, alpha, Aptrs, lda,
-                          Bptrs, ldb, beta, Cptrs, ldc, group_count, group_size)
-            else
-                $fname(handle(), transa, transb, m, n, k, alpha, Aptrs, lda,
-                          Bptrs, ldb, beta, Cptrs, ldc, group_count, group_size)
+            try
+                ## XXX: cublasXgemmGroupedBatched does not seem to support device pointers
+                cublasSetPointerMode_v2(handle(), CUBLAS_POINTER_MODE_HOST)
+
+                if CUBLAS.version() >= v"12.0"
+                    $fname_64(handle(), transa, transb, m, n, k, alpha, Aptrs, lda,
+                              Bptrs, ldb, beta, Cptrs, ldc, group_count, group_size)
+                else
+                    $fname(handle(), transa, transb, m, n, k, alpha, Aptrs, lda,
+                              Bptrs, ldb, beta, Cptrs, ldc, group_count, group_size)
+                end
+            finally
+                cublasSetPointerMode_v2(handle(), CUBLAS_POINTER_MODE_DEVICE)
             end
             unsafe_free!(Cptrs)
             unsafe_free!(Bptrs)
--- a/lib/cublas/wrappers.jl
+++ b/lib/cublas/wrappers.jl
@@ -352,7 +352,8 @@ for (fname, fname_64, elty, sty) in (
                       y::StridedCuVecOrDenseMat{$elty},
                       c::C,
                       s::S,
-                     ) where {C<:Union{Ref{Real}, Real, AbstractArray{Real}}, S<:Union{Ref{$sty}, $sty, AbstractArray{$sty}}}
+                     ) where {C<:Union{Ref{Real}, Real, AbstractArray{Real}},
+                              S<:Union{Ref{$sty}, $sty, AbstractArray{$sty}}}
             if CUBLAS.version() >= v"12.0"
                 $fname_64(handle(), n, x, stride(x, 1), y, stride(y, 1), c, s)
             else
@@ -371,10 +372,10 @@ for (fname, elty) in ((:cublasSrotg_v2, :Float32),
                      )
     @eval begin
         function rotg!(a::$elty, b::$elty)
-            c = Ref{real($elty)}(zero(real($elty)))
-            s = Ref{$elty}(zero($elty))
-            ref_a = Ref(a)
-            ref_b = Ref(b)
+            c = CuRef{real($elty)}()
+            s = CuRef{$elty}()
+            ref_a = CuRef(a)
+            ref_b = CuRef(b)
             $fname(handle(), ref_a, ref_b, c, s)
             ref_a[], ref_b[], c[], s[]
         end
@@ -389,7 +390,7 @@ for (fname, fname_64, elty) in ((:cublasSrotm_v2, :cublasSrotm_v2_64, :Float32),
         function rotm!(n::Integer,
                        x::StridedCuVecOrDenseMat{$elty},
                        y::StridedCuVecOrDenseMat{$elty},
-                       param::AbstractVector{$elty})
+                       param::CuVector{$elty})
             if CUBLAS.version() >= v"12.0"
                 $fname_64(handle(), n, x, stride(x, 1), y, stride(y, 1), param)
             else
@@ -408,13 +409,13 @@ for (fname, elty) in ((:cublasSrotmg_v2, :Float32),
                         d2::$elty,
                         x::$elty,
                         y::$elty,
-                        param::AbstractVector{$elty})
-            ref_d1 = Ref(d1)
-            ref_d2 = Ref(d2)
-            ref_x  = Ref(x)
-            ref_y  = Ref(y)
+                        param::CuVector{$elty})
+            ref_d1 = CuRef(d1)
+            ref_d2 = CuRef(d2)
+            ref_x  = CuRef(x)
+            ref_y  = CuRef(y)
             $fname(handle(), ref_d1, ref_d2, ref_x, ref_y, param)
-            ref_d1[], ref_d2[], ref_x[], ref_y[], param 
+            ref_d1[], ref_d2[], ref_x[], ref_y[], param
         end
     end
 end
@@ -497,7 +498,7 @@ for fname in (:iamax, :iamin)
         function $fname(n::Integer, dx::StridedCuVecOrDenseMat)
             result_type = CUBLAS.version() >= v"12.0" ? Int64 : Cint
             result = CuRef{result_type}()
-            $fname(n, dx, gpu_result)
+            $fname(n, dx, result)
             return result[]
         end
         $fname(dx::StridedCuVecOrDenseMat) = $fname(length(dx), dx)
--- a/lib/cublas/wrappers.jl
+++ b/lib/cublas/wrappers.jl
@@ -252,7 +252,7 @@ for (fname, fname_64, elty, ret_type) in ((:cublasDnrm2_v2, :cublasDnrm2_v2_64, 
 end
 
 nrm2(x::StridedCuVecOrDenseMat) = nrm2(length(x), x)
-nrm2(x::StridedCuVecOrDenseMat, result::CuVector) = nrm2(length(x), x, result)
+nrm2(x::StridedCuVecOrDenseMat, result) = nrm2(length(x), x, result)
 
 function nrm2(n::Integer, x::StridedCuVecOrDenseMat{Float16}, result)
     cublasNrm2Ex(handle(), n, x, Float16, stride(x, 1), result, Float16, Float32)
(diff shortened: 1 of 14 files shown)


kshyatt:
I can also add some more `@eval` blocks to try to cut down on the repetitive fallback logic

kshyatt:
Sample speedup:

```julia
julia> using CUDA, CUDA.CUBLAS, LinearAlgebra;

julia> n = Int(2^26);

julia> X = CUDA.rand(Float64, n);

julia> res = CuRef{Float64}(0.0);

# do some precompilation runs first

julia> @time CUBLAS.nrm2(n, X, res);
  0.000104 seconds (18 allocations: 288 bytes)

julia> @time CUBLAS.nrm2(n, X);
  0.001564 seconds (73 allocations: 3.094 KiB)

````

File: lib/cublas/wrappers.jl [hunk 1]
```diff
@@ -126,17 +126,29 @@ for (fname, fname_64, elty) in ((:cublasDscal_v2, :cublasDscal_v2_64, :Float64),
...
+    x
+end
+function scal!(n::Integer, alpha::CuRefArray{Float32, CuVector{Float32, DeviceMemory}}, x::StridedCuVecOrDenseMat{Float16})
+    cublasScalEx(handle(), n, alpha, Float32, x, Float16, stride(x, 1), Float32)
+    return x
+end
 function scal!(n::Integer, alpha::Number, x::StridedCuVecOrDenseMat{Float16})
-    α = convert(Float32, alpha)
-    cublasScalEx(handle(), n, Ref{Float32}(α), Float32, x, Float16, stride(x, 1), Float32)
+    α = CuRef{Float32}( convert(Float32, alpha) )
```
maleadt: We should improve `CuRef` so that it can be constructed identically to `Ref`. `Ref{T}(x)` doing an implicit `convert` is pretty convenient.

kshyatt:
Yeah, should one of us open an issue?

kshyatt:
Is the test failure something I've done? Seems GPUArrays related

kshyatt:
OK, I think this is ready for review!

Jutho:
I am not qualified to review, but certainly interested in the outcome. Will the non-blocking methods only accept `CuRef` objects for the scalar input or output quantities, or also zero-dimensional arrays (i.e. `CuArray{T,0}`)?

kshyatt:
For now only CuRef but these are easy to create (it’s exported by CUDA.jl).
I think one can also create them without a copy from a regular CuArray?

kshyatt:
You can create a `CuRefArray{T}` where `T` is some element type from a single element `CuVector`. In fact, `CuRef` itself does this under the hood.

lib/cublas/wrappers.jl
```diff
@@ -126,17 +126,29 @@ for (fname, fname_64, elty) in ((:cublasDscal_v2, :cublasDscal_v2_64, :Float64),
         end
     end
 end
+function scal!(n::Integer, alpha::Number, x::StridedCuVecOrDenseMat{T}) where {T}
+    α = convert(T, alpha)
+    gpu_α = CuRef{T}(α)
```
maleadt: The converts can go, `CuRef` does that for you:

```
julia> CuRef{Float32}(1)
CUDA.CuRefArray{Float32, CuArray{Float32, 1, CUDA.DeviceMemory}}(Float32[1.0], 1)
```

lib/cublas/wrappers.jl
```diff
@@ -126,17 +126,29 @@ for (fname, fname_64, elty) in ((:cublasDscal_v2, :cublasDscal_v2_64, :Float64),
...
+    scal!(n, gpu_α, x)
+    synchronize()
```
maleadt: Why the synchronization? The only way to see the changes by this call is to fetch memory, which is a synchronizing operation.

lib/cublas/wrappers.jl
```diff
@@ -115,7 +115,7 @@ for (fname, fname_64, elty) in ((:cublasDscal_v2, :cublasDscal_v2_64, :Float64),
                                 (:cublasCscal_v2, :cublasCscal_v2_64, :ComplexF32))
     @eval begin
         function scal!(n::Integer,
-                       alpha::Number,
+                alpha::CuRefArray{$elty, CuVector{$elty, DeviceMemory}},
                        x::StridedCuVecOrDenseMat{$elty})
```
maleadt: I know this is because the formatter only ran on the diff, but I'm no fan of partially aligned signatures.

maleadt (commented):
I wonder if we should also improve `CuRef` to initialize its memory by calling `fill` instead of `memcpy`: When calling `memcpy`, the copy likely won't be truly asynchronous (that would require pinned memory). But if we call `fill`, which should be possible for most scalars, the argument is passed by value and I think the call will complete asynchronously.
Something to investigate!

maleadt:
> Something to investigate!

https://github.com/JuliaGPU/CUDA.jl/pull/2625

lib/cublas/wrappers.jl
kshyatt: Grrr stupid formatter, I agree

lib/cublas/wrappers.jl
kshyatt: For `scal!` and other functions which don't return a scalar result, I added this to keep the previous behaviour (so that the entire call is synchronous). I'll remove the sync for things like `nrm2!` that return a scalar which is copied back anyway.

maleadt:
CI failures seem relevant.

Feel free to ignore the formatter; I made it less spammy 😉

kshyatt:
I really do not know what is up with the 1.11 failure, it looks `alloc_cache` related?

maleadt:
Rebase to get rid of CI failures?

kshyatt:
Yep, next on my to do list

kshyatt:
Gotta admit I'm a bit mystified here as I cannot reproduce these `trmm` faliures locally.

If I run *only* the `libraries/cublas` tests or even just `libraries` using the `runtests.jl` argument support, everything succeeds locally. If I run the full test suite, I start seeing intermittent illegal access errors/incorrect results in `syr2k!`. Weird!

lib/cublas/wrappers.jl
maleadt: > For `scal!` and other functions which don't return a scalar result, I added this to keep the previous behaviour (so that the entire call is synchronous).

That kind of defeats the purpose of this change, no? Since this is operating on GPU memory, which can't be fetched without synchronizing anyway, I don't see why we should keep the previous synchronous behavior. Making it asynchronous shouldn't break anything.

maleadt: In fact, this method should have been asynchronous before (as it doesn't return a scalar, https://docs.nvidia.com/cuda/cublas/#scalar-parameters), so synchronizing now seems like a regression.

lib/cublas/wrappers.jl
```diff
@@ -126,18 +126,29 @@ for (fname, fname_64, elty) in ((:cublasDscal_v2, :cublasDscal_v2_64, :Float64),
...
     return x
 end
 # specific variants in case x is complex and alpha is real
 for (fname, fname_64, elty, celty) in ((:cublasCsscal_v2, :cublasCsscal_v2_64, :Float32, :ComplexF32),
                                        (:cublasZdscal_v2, :cublasZdscal_v2_64, :Float64, :ComplexF64))
     @eval begin
         function scal!(n::Integer,
-                       alpha::$elty,
-                       x::StridedCuVecOrDenseMat{$celty})
+                       alpha::CuRefArray{$elty, CuVector{$elty, M}},
```
maleadt: Any reason you're typing these so specifically? I"d just use `alpha::Ref{$elty}` (`CuRefArray` isn't `<: CuRef` because we can't have `CuRef` be abstract and work with `ccall`; `Ref` is special, see `jl_is_abstract_ref_type`).

maleadt: I simplified it, feel free to revert if there was a specific reason.

maleadt:
We could simplify the implementation here by switching the CUBLAS signatures from `RefOrCuRef` to simply `CuRef` (which IIUC is the only legal option now that we switched to DEVICE_POINTER mode), so that passing a scalar is automatically converted to a CuRef:

```patch
--- a/lib/cublas/wrappers.jl
+++ b/lib/cublas/wrappers.jl
@@ -530,10 +530,10 @@ for (fname, fname_64, elty) in ((:cublasDgemv_v2, :cublasDgemv_v2_64, :Float64),
                                 (:cublasCgemv_v2, :cublasCgemv_v2_64, :ComplexF32))
     @eval begin
         function gemv!(trans::Char,
-                       alpha::Ref{$elty},
+                       alpha,
                        A::StridedCuMatrix{$elty},
                        x::StridedCuVector{$elty},
-                       beta::Ref{$elty},
+                       beta,
                        y::StridedCuVector{$elty})
             # handle trans
             m,n = size(A)
@@ -552,23 +552,10 @@ for (fname, fname_64, elty) in ((:cublasDgemv_v2, :cublasDgemv_v2_64, :Float64),
         end
     end
 end
-function gemv!(trans::Char, alpha::Number, A::StridedCuMatrix{T}, x::StridedCuVector{T}, beta::Number, y::StridedCuVector{T}) where {T}
-    gpu_α = CuRef{T}(alpha)
-    gpu_β = CuRef{T}(beta)
-    y = gemv!(trans, gpu_α, A, x, gpu_β, y)
-    synchronize()
-    return y
-end
-function gemv(trans::Char, alpha::Ref{T}, A::StridedCuMatrix{T}, x::StridedCuVector{T}) where {T}
-    return gemv!(trans, alpha, A, x, CuRef{T}(zero(T)), similar(x, size(A, (trans == 'N' ? 1 : 2))))
-end
-function gemv(trans::Char, alpha::Number, A::StridedCuMatrix{T}, x::StridedCuVector{T}) where T
+gemv(trans::Char, alpha, A::StridedCuMatrix{T}, x::StridedCuVector{T}) where {T} =
     gemv!(trans, alpha, A, x, zero(T), similar(x, size(A, (trans == 'N' ? 1 : 2))))
-end
-# should this be async?
-function gemv(trans::Char, A::StridedCuMatrix{T}, x::StridedCuVector{T}) where T
+gemv(trans::Char, A::StridedCuMatrix{T}, x::StridedCuVector{T}) where {T} =
     gemv!(trans, one(T), A, x, zero(T), similar(x, T, size(A, (trans == 'N' ? 1 : 2))))
-end

 for (fname, fname_64, eltyin, eltyout, eltyconst) in (
         (:cublasDgemvBatched, :cublasDgemvBatched_64, :Float64, :Float64, :Float64),
```

etc for the others. Thoughts?

kshyatt:
> We could simplify the implementation here by switching the CUBLAS signatures from RefOrCuRef to simply CuRef (which IIUC is the only legal option now that we switched to DEVICE_POINTER mode), so that passing a scalar is automatically converted to a CuRef

Yeah, that makes sense. I was biased towards keeping the changes more "conservative" and perhaps changing `handle()` to allow users to set the pointer mode, but if we want to go "device side by default" then this would make life easier.

kshyatt:
In the latest commit, I applied @maleadt 's suggestions widely. This really shortens the diff and tests passed locally. @Jutho , this should also allow your use-case of passing `CuArray{T, 0}` for the scalars as we have an auto-conversion of those to `CuRef`. If this run passes, I'll update the wrapper-generator signatures too.

kshyatt:
:rotating_light: FOUND THE CAUSE OF THE TEST FAILURE :rotating_light: 

[This block](https://github.com/JuliaGPU/CUDA.jl/blob/86b255e2f7f6b824cb63444ce40a4e4c6108e1d2/test/core/cudadrv.jl#L18) in `test/core/cudadrv.jl` causes the CUBLAS device side pointer tests to fail. I think the solution would be to set the device mode to exclusive when we set the pointer mode to device side. I'll try that and see if it resolves things.

maleadt:
Pushed some simplifications and clean-ups (RefOrCuRef can go after converting all scalars, I think, and I went back to how `pointer.jl` was implemented before this PR but moving the `@gcsafe_ccall` workaround from `RefOrCuRef` to `CuRef`).

Going to mark the PR as draft though as there's some obvious issues / debugging code in here.

kshyatt:
Tests passed! Pushed a fix for the wrapper generator to have `cublasXt` functions use `Ref` rather than `CuRef` for their scalars - since the `Xt` methods are multi-GPU, and `cublasXt` doesn't have this pointer mode business, I think we should keep the CPU-side scalar for now.

maleadt:
Pushed some simplificatons, e.g., supporting `CuRef(x)[]`. The new wrappers from https://github.com/JuliaGPU/CUDA.jl/pull/2642 also broke this PR, in part because there's still many `PtrOrCuPtr`s in there that are now illegal.

Finally, I don't feel like we should simply remove the group batched functionality. Since the handle is task-local anyway, I'd rather we temporarily switch pointer modes for these API calls and keep the functionality working.

test/runtests.jl
```diff
@@ -402,6 +396,12 @@ try
                     else
                         print_testworker_stats(test, wrkr, resp)
                     end
+
+                    # resetting the context breaks certain CUDA libraries,
+                    # so spawn a new worker when the test did so
+                    if test in ["core/initialization", "core/cudadrv"]
+                        p = recycle_worker(p)
+                    end
```
kshyatt, line 404: Ah cool I completely missed that this function existed

kshyatt:
Are we still "needs changes" on this or can it be squash-merged now that it passed CI?

maleadt:
I think this is fine for now.

Issue #1760, opened 2023-02-08
Mix precision sparse mul is not dispatched correctly


Roger-luo:
**Describe the bug**

The following should be working

```julia
using CUDA
using Adapt
using CUDA.CUSPARSE
using SparseArrays
using LinearAlgebra
CUDA.allowscalar(false)

A = rand(ComplexF64, 10)
B = sprand(Float64, 10, 10, 0.5)
C = rand(ComplexF64, 10)

dA = adapt(CuArray, A)
dB = adapt(CuArray, B)
dC = adapt(CuArray, C)
mul!(A, B, C)
mul!(dA, dB, dC)
```

and it should be dispatched to

```julia
julia> @which mul!(dA, dB, dC, 0.1, 0.1)
mul!(C::CuArray{Complex{T}, 1}, A::CuSparseMatrix{T}, B::CuArray{Complex{T}, 1}, alpha::Number, beta::Number) where T<:Union{Float16, Float32, Float64, ComplexF32, ComplexF64} in CUDA.CUSPARSE at /home/ubuntu/.julia/packages/CUDA/BbliS/lib/cusparse/interfaces.jl:44
```

But now in CUDA 4.x, this is calling the fallback implementation in Base, I'm not sure what has been changed in CUSPARSE since #1475, I think this should be an easy fix somewhere. I can submit a PR later.

kshyatt:
Was able to repro and fix this, PR coming soon after I check whether it's also a problem for mat-mat mul.

PR #2651, opened 2025-02-13
Re-enable mixed precision sparse mv

kshyatt:
Should address https://github.com/JuliaGPU/CUDA.jl/issues/1760

### DIFF
--- a/lib/cusparse/interfaces.jl
+++ b/lib/cusparse/interfaces.jl
@@ -62,19 +62,19 @@ op_wrappers = ((identity, T -> 'N', identity),
                (T -> :(HermOrSym{T, <:$T}), T -> 'N', A -> :(parent($A))))
 
 # legacy methods with final MulAddMul argument
-LinearAlgebra.generic_matvecmul!(C::CuVector{T}, tA::AbstractChar, A::CuSparseMatrix{T}, B::DenseCuVector{T}, _add::MulAddMul) where {T <: Union{Float16, ComplexF16, BlasFloat}} =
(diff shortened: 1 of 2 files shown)


amontoison (approved):
We could also do the same thing for `generic_matmatmul!`.

kshyatt:
@amontoison looking at the docs for the various generic `mm` methods, I think none of them support mixed precision for `BlasFloat` combinations for any non ELL storage format? Weirdly it seems to be only sparse mat * dense vec that does.

amontoison:
I checked the code of `mm!` in CUSPARSE and we need to add another parametric type `TA`:
https://github.com/JuliaGPU/CUDA.jl/blob/master/lib/cusparse/generic.jl#L157-L278

kshyatt:
Right but `SpMM` does NOT support `ComplexF32`/`Float32` element type combinations for CSR/CSC/COO matrices

amontoison:
Sorry I didn't know that you wanted mixed precision with simple / double precision.
Only low precision is supported in mixed mode for `mm!`:
Capture d’écran du 2025-02-12 18-15-27

kshyatt:
Yeah, it's a bit odd because mixed real/complex is supported for CSC/CSR/COO for sparse m * dense v. That's why I didn't handle `generic_matmul!` in this PR...

amontoison:
Ok LGTM, I learned something :)
I thought that `mv!` and `mm!` supported the same mixed-precision modes.

kshyatt:
It would certainly make sense if they did! Unfortunately it seems to only be the case for the ELL storage format...

PR #2660, opened 2025-02-15
Add tests for gemmEx in fast math mode

kshyatt:


maleadt (changes requested):
Probably needs to be put in a `try`/`finally`; we wouldn't want the bad math mode to leak when an error occurs.

Or, put the tests in a separate `@async |> fetch`, where it's fine to set the math mode and forget about it.

maleadt (approved):

PR #2669, opened 2025-02-20
Exclude lib*jl from coverage also for CUSTATEVEC, CUTENSOR, and CUTENSORNET

kshyatt:


kshyatt:
Enzyme error doesn't look like it could be related?

PR #2677, opened 2025-03-01
More tests, better errors, more exclusions for CUSPARSE

kshyatt:


amontoison (approved):

kshyatt:
@amontoison this block starting [here](https://app.codecov.io/gh/JuliaGPU/CUDA.jl/blob/ksh%2Fcusparse_again/lib%2Fcusparse%2Fconversions.jl#L280) seems unreachable? Do you have an example test that can hit it? If not, should we just delete?

PR #2707, opened 2025-03-20
More tests and bugfixes for CUSOLVER

kshyatt:
So it appears we have a duplication of `inv` for triangulars - if the matrix is strided, CUSOLVER is used, but if it's dense, CUBLAS is used. Not sure we want to keep that but for now this finally tests and fixes the CUSOLVER one and addresses another bug in `Symmetrix \ CuMat`.

### DIFF
--- a/lib/cusolver/dense_generic.jl
+++ b/lib/cusolver/dense_generic.jl
@@ -122,7 +122,7 @@ function Xgeqrf!(A::StridedCuMatrix{T}) where {T <: BlasFloat}
 end
 
 # Xsytrs
-function sytrs!(uplo::Char, A::StridedCuMatrix{T}, p::CuVector{Int64}, B::StridedCuMatrix{T}) where {T <: BlasFloat}
+function sytrs!(uplo::Char, A::StridedCuMatrix{T}, p::CuVector{Int64}, B::StridedCuVecOrMat{T}) where {T <: BlasFloat}
     chkuplo(uplo)
     n = checksquare(A)
     nrhs = size(B, 2)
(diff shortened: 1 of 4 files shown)


amontoison:
What is the method of CUBLAS that is used?
Is it `trsm` used with `I` as right-hand side?

maleadt:
I think it dispatches from `ldiv!` through `generic_trimatdiv!` to `trsv!`.

It seems silly to keep both definitions, though. Can we benchmark them and select one? I would expect the specialized CUSOLVER method to perform better, but from a quick look it seems like the inverse (hah).

lib/cusolver/dense_generic.jl
```diff
@@ -122,7 +122,7 @@ function Xgeqrf!(A::StridedCuMatrix{T}) where {T <: BlasFloat}
 end
 
 # Xsytrs
-function sytrs!(uplo::Char, A::StridedCuMatrix{T}, p::CuVector{Int64}, B::StridedCuMatrix{T}) where {T <: BlasFloat}
+function sytrs!(uplo::Char, A::StridedCuMatrix{T}, p::CuVector{Int64}, B::StridedCuVecOrMat{T}) where {T <: BlasFloat}
```
amontoison, line 125: @kshyatt
A test will be also great to ensure that we don't need to adapt anymore for `CuVector` in sytrs!.

kshyatt: Sorry, can you clarify what test you'd like a little? (I haven't had coffee yet). We have a test for `CuVector` `B` in:
`@test Array(Symmetric(d_A) \ d_b) ≈ (Af \ bf)` down below, no?

amontoison: The tests of sytrs is in this file:
https://github.com/JuliaGPU/CUDA.jl/blob/master/test/libraries/cusolver/dense_generic.jl#L94

You are right, `Array(Symmetric(d_A) \ d_b) ≈ (Af \ bf)` should target `sytrs` but it is not a direct test on it. Also CUSOLVER has two API (legacy and 64-bit integer) and we start to have two versions (Xgetrf and getrf) for each LAPACK routine.

kshyatt: OK, test added!

amontoison (approved):

maleadt:
> So it appears we have a duplication of `inv` for triangulars - if the matrix is strided, CUSOLVER is used, but if it's dense, CUBLAS is used.

So we're still keeping both versions?

PR #2744, opened 2025-04-14
More tests and better error type for cusparse generic

kshyatt:


lib/cusparse/generic.jl
```diff
@@ -82,8 +82,6 @@ function densetosparse(A::CuMatrix{T}, fmt::Symbol, index::SparseChar, algo::cus
             nzVal = CuVector{T}(undef, nnzB[])
             B = CuSparseMatrixCSC{T, Cint}(colPtr, rowVal, nzVal, (m,n))
             cusparseCscSetPointers(desc_sparse, B.colPtr, B.rowVal, B.nzVal)
-        else
```
kshyatt, line 85: This I think we can remove because it should be caught higher up in the function

maleadt (approved):

PR #2789, opened 2025-05-28
Support JLD2 for (de)serialization

lassepe:
Currently, serializing a `CUDA` array with `JLD2` will silently fail because it serializes a pointer that is no longer valid when deserializing. This PR adds an extension for `JLD2` that handles this more gracefully by hooking into the [`JLD2.writeas` mechanism](https://juliaio.github.io/JLD2.jl/dev/customserialization/).


lassepe:
Buildkite failure seems unrelated.

lassepe:
The more I think about this, maybe this kind of overloading should happen directly on `GPUArraysCore.AbstractGPUArray`?

lassepe:
I will try to shoot for a more general solution directly in JLD2: https://github.com/JuliaIO/JLD2.jl/issues/647
If those efforts result in a dead end, I will re-open this PR for `GPUArraysCore.jl`.

PR #2797, opened 2025-06-23
Better error messages and tests for sm2

kshyatt:
We can get rid of all this once we drop support for CUSPARSE 11 but in the meantime it's nice to have better error messages


amontoison (approved):

kshyatt:
Test fails don't look related to me? CI is very flaky right now

PR #2825, opened 2025-07-26
Trial support for thread-block clusters

kshyatt:
Trying to attack https://github.com/JuliaGPU/CUDA.jl/issues/1989

This should allow users to submit a `clusters=` kwarg at kernel launch, which will use `cuLaunchKernelEx` on supported hardware and throw an error if `clusters > 1` on any dimension on non-supported hardware. There are some tests but probably more can be done. I did wrap the necessary PTX instructions for detecting the cluster index and size within a kernel as well.

### DIFF
--- a/lib/cudadrv/execution.jl
+++ b/lib/cudadrv/execution.jl
@@ -44,11 +44,14 @@ end
 
 """
     launch(f::CuFunction; args...; blocks::CuDim=1, threads::CuDim=1,
-           cooperative=false, shmem=0, stream=stream())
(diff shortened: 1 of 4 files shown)


kshyatt:
Still needs support for:
- [] `cluster.sync`
- [] `barrier.arrive`
- [] `barrier.wait`
- [] `query_shared_rank`
- [] `map_shared_rank`

kshyatt:
I think we can close this, right? @eschnett's PR should have superseded it

eschnett:
Yes. Thanks for getting this started!

PR #2828, opened 2025-07-30
Make `resize!` run faster

huiyuxie:
I need `resize!` function run faster when it is called frequently.

I'm not sure about whether 2 is a good resize factor but I think it might be a reasonable number. And I do the benchmarks based on the assumption that the resize length is uniformly distributed within a range.
<details>
<summary>Click for benchmark script</summary>

```julia
using CUDA
using BenchmarkTools
using Random

# we assume resize length is uniformly distributed within a range

# case 1: range 1 to 100
a1 = CUDA.rand(50)
function old1(a::CuArray, rng)
    size = rand(rng, 1:100)
    CUDA.resize!(a, size)
end
function new1(a::CuArray, rng)
    size = rand(rng, 1:100)
    CUDA.new_resize!(a, size)
end
seed = 1
rng1a = MersenneTwister(seed)
ben1a = @benchmark CUDA.@sync old1(a1, rng1a)
rng1b = MersenneTwister(seed)
ben1b = @benchmark CUDA.@sync new1(a1, rng1b)

# case 2: range 1 to 1000
a2 = CUDA.rand(500)
function old2(a::CuArray, rng)
    size = rand(rng, 1:1000)
    CUDA.resize!(a, size)
end
function new2(a::CuArray, rng)
    size = rand(rng, 1:1000)
    CUDA.new_resize!(a, size)
end
seed = 12
rng2a = MersenneTwister(seed)
ben2a = @benchmark CUDA.@sync old2(a2, rng2a)
rng2b = MersenneTwister(seed)
ben2b = @benchmark CUDA.@sync new2(a2, rng2b)

# case 3: range 1 to 10000
a3 = CUDA.rand(5000)
function old3(a::CuArray, rng)
    size = rand(rng, 1:10000)
    CUDA.resize!(a, size)
end
function new3(a::CuArray, rng)
    size = rand(rng, 1:10000)
    CUDA.new_resize!(a, size)
end
seed = 123
rng3a = MersenneTwister(seed)
ben3a = @benchmark CUDA.@sync old3(a3, rng3a)
rng3b = MersenneTwister(seed)
ben3b = @benchmark CUDA.@sync new3(a3, rng3b)

# case 4: range 1 to 100000
a4 = CUDA.rand(50000)
function old4(a::CuArray, rng)
    size = rand(rng, 1:100000)
    CUDA.resize!(a, size)
end
function new4(a::CuArray, rng)
    size = rand(rng, 1:100000)
    CUDA.new_resize!(a, size)
end
seed = 1234
rng4a = MersenneTwister(seed)
ben4a = @benchmark CUDA.@sync old4(a4, rng4a)
rng4b = MersenneTwister(seed)
ben4b = @benchmark CUDA.@sync new4(a4, rng4b)

# case 5: range 1 to 1000000
a5 = CUDA.rand(500000)
function old5(a::CuArray, rng)
    size = rand(rng, 1:1000000)
    CUDA.resize!(a, size)
end
function new5(a::CuArray, rng)
    size = rand(rng, 1:1000000)
    CUDA.new_resize!(a, size)
end
seed = 12345
rng5a = MersenneTwister(seed)
ben5a = @benchmark CUDA.@sync old5(a5, rng5a)
rng5b = MersenneTwister(seed)
ben5b = @benchmark CUDA.@sync new5(a5, rng5b)
```
</details> 

<details>
<summary>Click for benchmark result</summary>

```julia
# range 1 to 100

# old
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  400.000 ns …  2.403 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):      19.800 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):    27.475 μs ± 41.275 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▂              ██▅▂▂▂▁      ▁▅▅▂▁▁    ▄▃▂▁                   ▂
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁█████████▇██████████▇███████████▇▆▅▅▅▄▆▅▅▅▅▃▄ █
  400 ns        Histogram: log(frequency) by time      70.7 μs <

# new
BenchmarkTools.Trial: 10000 samples with 3 evaluations per sample.
 Range (min … max):  366.667 ns … 897.633 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):      17.900 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):    17.700 μs ±  17.207 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

         ▂      █      ▆
  ▃▁▁▁▁▁▂█▃▂▂▂▂▃█▅▃▃▂▃▄█▇▄▃▃▃▃▄▄▃▂▃▃▃▃▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂ ▃
  367 ns           Histogram: frequency by time         51.2 μs <

```
```julia
# range 1 to 1000

# old
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  400.000 ns …  1.352 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):      20.000 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):    28.770 μs ± 43.310 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

             ██▆▄▄▃▂▁▁▁▂▅▄▃▂▁▁▁▄▁▁ ▁▁▁                         ▂
  ▆▁▁▁▁▁▁▁▁▁▁██████████████████████████▇▆▇▆▇▆▇▆▆▆▅▅▅▅▅▆▅▄▆▁▅▃▄ █
  400 ns        Histogram: log(frequency) by time      93.4 μs <

# new
BenchmarkTools.Trial: 10000 samples with 6 evaluations per sample.
 Range (min … max):   3.650 μs … 324.167 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     19.800 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   22.643 μs ±  16.182 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

       ▃▂█▃▆▁▄
  ▂▃▄▆▆███████▇█▅▄▅▃▃▂▃▃▃▂▂▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▂ ▃
  3.65 μs         Histogram: frequency by time          107 μs <

 Memory estimate: 85 bytes, allocs estimate: 4.
```
```julia
# range 1 to 10000

# old
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  500.000 ns …  1.365 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):      21.200 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):    28.338 μs ± 44.612 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

                █▇▅▅▃▃▂▂▂▂▁▁▄▄▃▂▁▁  ▁▁▁ ▁                      ▂
  ▃▁▁▁▁▁▁▁▁▁▁▁▁▁███████████████████▇█████▇▇▇▆▆▆▅▆▅▃▃▆▃▅▅▁▃▅▃▄▅ █
  500 ns        Histogram: log(frequency) by time        81 μs <

# new
BenchmarkTools.Trial: 10000 samples with 3 evaluations per sample.
 Range (min … max):  366.667 ns … 551.667 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):      19.367 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):    20.171 μs ±  16.419 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

         ▃      █▁     ▆▁
  ▃▂▁▁▁▁▂█▃▂▂▂▂▃██▄▃▃▃▅██▆▄▃▃▄▆▅▄▃▃▃▄▃▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▃▂▂▂▂▂▂▂▂▂ ▃
  367 ns           Histogram: frequency by time         54.8 μs <

```
```julia
# range 1 to 100000

# old
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  16.100 μs … 771.100 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     20.300 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   26.895 μs ±  31.561 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▅██▇▅▄▄▂▂▃▄▃▃▂▁▁▁ ▁                                          ▂
  ███████████████████████▇▇▇▇▇█▇▇▇▇▇▇█▇▇▆▅▅▅▅▄▆▆▄▄▅▅▄▄▃▂▄▄▃▄▂▄ █
  16.1 μs       Histogram: log(frequency) by time       105 μs <

 Memory estimate: 512 bytes, allocs estimate: 24.
# new
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  300.000 ns … 805.600 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):      17.700 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):    17.344 μs ±  28.719 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █                  ▃▇▇▇▆▅▃▂▁      ▁▂▃▄▃▃▂▁                    ▂
  █▆▃▄▁▄▄▃▁▁▁▁▁▁▁▁▁▁▁████████████▇▇███████████▆▇▆▆▆▆▇███▇▇▇▇▅▅▅ █
  300 ns        Histogram: log(frequency) by time       49.9 μs <

```
```julia
# range 1 to 1000000

# old
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  11.600 μs … 236.500 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     19.000 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   21.556 μs ±  11.217 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

   ▄▇█▇▄  ▃▇█▇▅▃▂
  ▆█████▅▄███████▇▆▄▃▃▂▂▂▂▂▂▂▂▂▁▂▂▂▂▂▂▂▃▃▂▂▂▂▂▂▂▂▂▂▂▁▁▁▁▁▂▁▁▁▁ ▃
  11.6 μs         Histogram: frequency by time         53.6 μs <

 Memory estimate: 512 bytes, allocs estimate: 24.
# new
BenchmarkTools.Trial: 10000 samples with 4 evaluations per sample.
 Range (min … max):  375.000 ns …  1.725 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):      15.550 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):    16.869 μs ± 23.618 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

               ▁▃▆▆▆▆▇▇▇██▇▆▅▅▄▄▂▁▁
  ▃▁▁▁▂▄▄▆▄▅▇▆█████████████████████▇▇▅▆▅▅▄▄▄▄▄▃▃▃▃▂▂▃▂▂▂▂▂▂▂▂▂ ▅
  375 ns          Histogram: frequency by time         40.1 μs <

```
</details> 

But it is still not as fast as CPU `resize!` function. 

Do you have any other suggestions to make it run faster (like any other good resize factor)? And do we need to benchmark it based on other assumptions (like the resize length linearly grows or shrinks with the times)? I guess the performance will not be good on some corner cases (like when we keep expanding the GPU array to less than double its length).

I separate the new `resize!` and the old `resize!` temporarily for better comparison.


huiyuxie:
@maleadt Please review. Thanks!

maleadt (changes requested):
Can you modify `resize!` instead of adding a `new_resize!`?

The change should also make use of the `maxsize` property of `CuArray`, which is there to allow additional data to be allocated without the dimensions of the array having to match. `resize` should be made aware of that, not doing anything when the needed size is smaller than `maxsize`.

In addition, IIUC you're using a growth factor of 2 now (ignoring resizes when shrinking by up to a half, resizing by 2 when requesting a larger array), which I think may be too aggressive for GPU arrays. For small arrays it's probably fine, but at some point (> 10MB?) we should probably use a fixed (1MB?) increment instead.

In addition,

huiyuxie:
> The change should also make use of the maxsize property of CuArray, which is there to allow additional data to be allocated without the dimensions of the array having to match.

👍

> which I think may be too aggressive for GPU arrays. For small arrays it's probably fine, but at some point (> 10MB?) we should probably use a fixed (1MB?) increment instead.

👍 but why do we choose 10MB and 1MB?

> In addition,

Do you have any other comments?

src/array.jl
```diff
@@ -876,42 +876,86 @@ Base.unsafe_convert(::Type{CuPtr{T}}, A::PermutedDimsArray) where {T} =
...
-  bufsize = if isbitstype(T)
-    maxsize
-  else
-    # type tag array past the data
-    maxsize + n
-  end
+  cap = A.maxsize ÷ aligned_sizeof(T)
+
+  # do nothing when new length is smaller than maxsize
+  if n > cap # n > length(A)
```
huiyuxie: > resize should be made aware of that, not doing anything when the needed size is smaller than maxsize

Is that really ok if we never shrink the GPU arrays?

huiyuxie:
Can you @maleadt  point me to the existing CI benchmark results? I could not find them. Thanks!

src/array.jl
```diff
@@ -876,42 +876,53 @@ Base.unsafe_convert(::Type{CuPtr{T}}, A::PermutedDimsArray) where {T} =
...
+    else 
+      len = max(n, 2 * length(A))
+    end
+
+    maxsize = len * aligned_sizeof(T)
+    bufsize = if isbitstype(T)
+        maxsize
+    else
+      # type tag array past the data
+      maxsize + len
```
huiyuxie, line 912: Do we have to cover the test for this line since https://app.codecov.io/gh/JuliaGPU/CUDA.jl/pull/2828?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=JuliaGPU?
But the original test also did not cover this line, I think.

huiyuxie:
@maleadt Could you please have a review if you get time? It's good if it can be merged ASAP.

huiyuxie:
Please review @maleadt. Thanks!

maleadt:
Sorry for the delay. I simplified the tests a little and added support for shrinking, as well as isbits-union arrays (although untested).



> but why do we choose 10MB and 1MB?

That was arbitrary. Would probably be useful to see what other projects do.

src/array.jl
maleadt: I added support for shrinking, it seems wasteful to keep precious GPU memory around of the user decides to give it back to the system.

maleadt:
CI failure related.

PR #2887, opened 2025-09-10
Support new cuquantum version 

kshyatt:
Marking this WIP for now because I need to check if there's additional API to support.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -99,15 +99,6 @@ steps:
               - "cuTENSOR"
               - "cuStateVec"
               - "cuTensorNet"
-          adjustments:
-            - with:
-                package: "cuStateVec"
-                cuda: "12.0"
-              soft_fail: true
-            - with:
-                package: "cuStateVec"
-                cuda: "13.0"
-              soft_fail: true
         plugins:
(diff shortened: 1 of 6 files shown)


kshyatt:
Going to try to use 25.11 instead which still works on CUDA 12, but has a bunch more features

kshyatt:
OK, I would rather just merge this now and I can work on adding more wrappers for a patch version in the future. The one thing I might wait on to merge this is fixing the `cuStateVec` errors. Thoughts?

kshyatt:
OK, so the tests were failing because the expectation computation returns immediately, and you have to synchronize to get the actual result. I changed the return a bit so that the references are returned, and users can choose when to synchronize.

maleadt (commented):
It's a bit ugly that we return a ref here yet you're only supposed to load from it after doing a synchronization. I wonder if we could generalize `CuRefValue` instead to support host memory, such that we benefit from its automatic synchronization? Or something dedicated to avoid a redundant `CuRef` -> `Ref` copy?

Just a thought, feel free to merge now as is, since you're probably the only user of these wrappers anyway.

kshyatt:
Yeah I very much agree it's grody, but I really do not want to silently synchronize in the wrapper if it can be avoided...

PR #2899, opened 2025-09-18
Updates for PrettyTables.jl v3

juliohm:
This is holding further updates in downstream ecosystems.

I removed the option `crop` because it was replaced by `fit_table_in_display_horizontally` and `fit_table_in_display_vertically`. Please feel free to adjust accordingly.

cc: @ronisbr


juliohm:
Side note: I cannot run tests locally. Julia is killed by the operating system to avoid memory shortage, and this has nothing to do with the changes in this PR.

ronisbr:
The function `time_highlighters` must also be changed. The text highlighters is now called `TextHighlighter`. The function must now also return a vector instead of a tuple.

ronisbr:
If I understood the code correctly, the `body_hlines` can be updated using:

```julia
table_format = TextTableFormat(; horizontal_lines_at_data_rows = trace_divisions)
```

and pass `table_format` to `pretty_table` function.

ronisbr:
The crop functionality can be updated using:

```julia
fit_table_in_display_vertically = false
fit_table_in_display_horizontally =
    if get(io, :is_pluto, false) || get(io, :jupyter, false)
        # Pluto.jl and IJulia.jl both indicate they want to limit output,
        # but they have scrollbars, so let's ignore that
        false
    elseif io isa Base.TTY || get(io, :limit, false)::Bool
        # crop horizonally to fit the terminal
        true
    else
        false
    end
```

Then, pass those two keywords to `pretty_table`.

juliohm:
@maleadt could you please help here? How to test the profiler's output? Also, do you wish to keep the `crop` and `body_hlines` options?

GregPlowman (commented):
formatters should return a Vector 
`formatters::Union{Nothing, Vector{Function}}`

https://ronisbr.github.io/PrettyTables.jl/stable/man/usage/#Formatters

I think lines 726, 811, 885

src/profile.jl
```diff
@@ -843,8 +831,9 @@ function Base.show(io::IO, results::ProfileResults)
 
             header = [summary_column_names[name] for name in names(df)]
             alignment = [name in ["name", "time_dist"] ? :l : :r for name in names(df)]
+            formatters = summary_formatter(df)
```
GregPlowman, line 834: formatters::Union{Nothing, Vector{Function}}

ronisbr:
> formatters should return a Vector

Yes! I missed that one! All formatters now must be inside a vector.

juliohm:
@ronisbr please feel free to suggest changes in this PR. I am happy to merge any suggestion. Also, feel free to start a fresh PR with a better approach.

@maleadt it would be really nice if you or someone else in the JuliaGPU org could review and merge this. This dependency on PrettyTables.jl v2 is locking various updates downstream.

kshyatt:
@ronisbr This is still failing CI. We asked for your help with an upgrade guide from v2 to v3 precisely so we could unblock downstream updates.

ronisbr:
Hi @kshyatt and @juliohm !

I provided some guidance how to update the API here. The problem is that I have no way to test them :( I can submit a PR, but it will be 100% blind.

kshyatt:
I think the update about `crop` was the last thing we needed. See https://github.com/JuliaGPU/CUDA.jl/pull/2891 - this passed locally for me. Thank you for the info!

ronisbr:
Awesome! I’m here to help if you need anything. If we’re at a standstill, we could chat in Slack and see if I can get SSH access to a machine with a Nvidia GPU. That way, I can help with the migration. However, I’ll need to wait until I get back to Brazil because I’m in a pretty different time zone right now.

kshyatt:
I think this can be closed as the other PR passed CI. Thanks again for the help!

juliohm:
Thank you!

PR #2929, opened 2025-10-07
cuSPARSE: Interface generic mv! for SparseMatrixBSR

amontoison:
- Support `CuSparseMatrixBSR` for generic sparse `mv!` added with CUDA `13.0.1`.
- I also removed also some dead code for CUDA < 12.0
- We should not call the "preprocess" routines for `mv!` and `mm!` because we don't have any high-level way to reuse the buffer / descriptor. We just do more work for nothing.


kshyatt:
CI failures are related

PR #2951, opened 2025-10-28
XsyevBatched! interface accepting 3D StridedCuArray

quantumsteve:
I noticed that the batched eigensolver function [CUSOLVER.heevjBatched!](https://github.com/JuliaGPU/CUDA.jl/blob/b0b484c5bf7e7a8342e0285e42c5e235aa252c32/lib/cusolver/dense.jl#L771) accepts a 3-dimensional StridedCuArray while the similar function CUSOLVER.XsyevBatched! accepts a 2-dimensional StridedCuMatrix.  I think having the same interface for both would be desirable.

### DIFF
--- a/lib/cusolver/dense_generic.jl
+++ b/lib/cusolver/dense_generic.jl
@@ -505,13 +505,13 @@ function Xgeev!(jobvl::Char, jobvr::Char, A::StridedCuMatrix{T}) where {T <: Bla
 end
 
 # XsyevBatched
-function XsyevBatched!(jobz::Char, uplo::Char, A::StridedCuArray{T,3}) where {T <: BlasFloat}
+function XsyevBatched!(jobz::Char, uplo::Char, A::StridedCuArray{T, 3}) where {T <: BlasFloat}
     CUSOLVER.version() < v"11.7.1" && throw(ErrorException("This operation is not supported by the current CUDA version."))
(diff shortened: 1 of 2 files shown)


kshyatt:
Test fail is due to https://github.com/JuliaGPU/CUDA.jl/issues/2971, unrelated to this PR

lib/cusolver/dense_generic.jl
```diff
@@ -505,6 +505,47 @@ function Xgeev!(jobvl::Char, jobvr::Char, A::StridedCuMatrix{T}) where {T <: Bla
 end
 
 # XsyevBatched
+function XsyevBatched!(jobz::Char, uplo::Char, A::StridedCuArray{T, 3}) where {T <: BlasFloat}
+    CUSOLVER.version() < v"11.7.1" && throw(ErrorException("This operation is not supported by the current CUDA version."))
```
kshyatt: Would be good to provide more information here -- what *is* the current version? What's the minimum?

kshyatt: I understand this copies the below but it would be nice to improve both ;)

kshyatt (approved):
LGTM although updating the error messages would be nice

kshyatt (approved):

PR #2960, opened 2025-11-05
write `CuArray` for clarity

raman-maker:
added `CuArray` for easy understanding.


maleadt:
I don't think this improves the sentence.

PR #2967, opened 2025-11-12
Add autodocs for submodules and libraries

gdalle:
Every time I check out cuSPARSE I have to go to the source file to figure out what's inside `CuSparseMatrixCSR`. This PR is a first step that displays all the docstrings of the libraries and submodules, so that once we actually write them they will appear on the website.
Feel free to reject or modify if you don't think that's a good idea, especially if it reveals/defines "public" API that you don't want to commit to.


kshyatt:
@gdalle you can put `[only docs]` in the commit message to avoid running the full ci (expensive!!!!) as shown [in the README](https://github.com/JuliaGPU/CUDA.jl?tab=readme-ov-file#note-for-contributors-about-ci-usage)

gdalle:
Oops my bad! I put it in the second commit, hope it is enough

gdalle:
It seems the test suite still ran despite the following commits including "[only docs]". Does this guideline only apply to the very first commit? Should I start a new PR?

kshyatt:
Could you squash and repush with only docs in the command line? It can be a
bit finicky

gdalle:
I tried but it seems to be running anyway? Should I open a new PR?

kshyatt:
I mean if the docs build succeeds, I'm happy to merge. We can just let it run for now tbh.

gdalle:
I'd love to learn what went wrong for next time though

gdalle:
Sorry I couldn't test the docs build locally, I don't have an NVIDIA GPU

gdalle:
Closed in favor of #2972

PR #2970, opened 2025-11-13
Add @inbounds to SparseArrays.nzrange for CuSparseDeviceMatrixCSC

Zinoex:
Add @inbounds to SparseArrays.nzrange(g::CuSparseDeviceMatrixCSC, col::Integer) to avoid bounds checking the `colPtr` (that was causing me much register spilling into local mem), consistent with the new SparseArrays functionality of GPUArrays.jl.


kshyatt:
Just fyi these types are probably going away soon -- I honestly don't remember if I had `@inbounds` on the GPUArrays.jl function 🙈

Zinoex:
I'll cross that bridge when I need to. The GPUArrays.jl sparse device arrays look to be a drop-in replacement, but I also have some quite complex kernels in IntervalMDP.jl operating on CuSparseMatrixCSC, and therefore, I expect to run into issues when transitioning - and I most certainly don't have the time to fix that atm.

maleadt (approved):

PR #2973, opened 2025-11-14
[CUSPARSE] Fix wrong method for `kron` with `Diagonal` and add support to `kron` with FillArrays

albertomercurio:
The previous implementation of `kron` between a CUSPARSE matrix and a `Diagonal` was wrong. It was assuming the diagonal to be the identity, which is not always the case. I have extended it to any general `Diagonal` matrix embedding a `CuVector`.

In order to support the `kron` with identity like matrices, I have added the extension to FillArrays.jl. Here the diagonal is filled by an unique value.

### DIFF
--- a/CITATION.bib
+++ b/CITATION.bib
@@ -19,5 +19,6 @@
   volume        = {132},
   pages         = {29--46},
   year          = {2019},
+  doi           = {https://doi.org/10.1016/j.advengsoft.2019.02.002},
   publisher     = {Elsevier}
 }


maleadt:
Replaces https://github.com/JuliaGPU/CUDA.jl/pull/2804?

albertomercurio:
Oh I wasn't aware of that PR.

I would say that both fix the bug. This PR also implements the `kron` with a true Identity (FillArrays.jl).

albertomercurio:
Hello, any thoughts on this?

maleadt:
These changes seem to cause test failures.

albertomercurio:
Ok I should have fixed the errors. There are some errors on Julia v1.12 and nightly related to `norm`, which should not be related to this PR.

albertomercurio:
Ok by updating the branch all the tests pass

PR #2975, opened 2025-11-17
Ensure diagm preserves eltype

kshyatt:
Previously you could lose precision by using `diagm`. Very bad!


kshyatt:
This PR also prevents a CPU->GPU copy

PR #2976, opened 2025-11-18
Remove old set of deps test

kshyatt:
This is completely blocking CI yet only appears on the old dependency test. No one has opened an issue about it, suggesting it's not affecting users (yet...). Until/unless someone runs up a flag, should we let this soft fail for now?

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -402,32 +402,32 @@ steps:
             build.message !~ /\[skip special\]/
         timeout_in_minutes: 45
 
-      - label: ":older_man: Old dependencies"
-        plugins:
-          - JuliaCI/julia#v1:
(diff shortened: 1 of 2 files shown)


maleadt (changes requested):
The run is there for a reason. If we can't fix it, I'd rather remove it than add yet another soft-fail.

Can you reproduce locally? Have you tried bumping compat bounds to see which upgrade causes the issue?

kshyatt:
I'll try to repro locally again, Resolver wasn't working for me earlier but today is a new day :)

kshyatt:
Couldn't repro locally but discussed on Slack that this specific `base/threading` set of tests is known to be flaky (see comment at the top of the file) so it may make sense to disable it for now

PR #2990, opened 2025-12-07
Fix undefined variable in memory_source error message

KaanKesginLW:
The fallback `memory_source(::Any)` method references an undefined variable `t` in its error message. This causes an `UndefVarError` instead of the intended helpful message when passing an unsupported type to `CuTexture`.

**Before:** `memory_source(::Any) = error("Unknown texture source $(typeof(t))")`
**After:** `memory_source(t::Any) = error("Unknown texture source $(typeof(t))")`


maleadt (approved):
Thanks!

PR #2998, opened 2025-12-16
Make Ahead-of-Time compilation feasible for packages using CUDA.jl

apozharski:
Currently, Ahead-of-Time (AOT) compilation for any package `using CUDA.jl` is impossible (with or without `--trim`). This is due to several things as far as I can tell:
1. Many methods in `CUDA.jl` which are visible in the global method table, `llvmcall` LLVM ir with `llvm.nvvm` intrinsics which therefore leak through to the AOT step which generates assembly, leading to compilation failures.
2. It seems odd that these methods are being compiled in but it seems the common denominator is that even though no `MethodInstance`s exist for the methods, the methods are concrete (unsure if this is the correct Julia terminology). This is because they either are defined to take concrete types (for example the `clock` methods) or no arguments at all.
3. There is another issue caused specifically by the the wrapper in `libcudadevrt.jl`. The "concrete" methods in that file contain `ccall`s to external functions which do not exist until they are dynamically loaded.  
4. Unfortunately even fixing all of this in the current state of the PR, there seems to be calls to (at least) `gpu_malloc`, `gpu_signal_exception`, and `gpu_report_oom` in the produced llvm ir. These functions don't exist as they would come from the runtime compilation of the runtime. This is likely a problem more fixable in `GPUCompiler.jl`, but maybe @vchuravy or @maleadt you have a better understanding of that.

The relevant issues and other PRs:
- The conversation here: https://github.com/JuliaGPU/GPUCompiler.jl/issues/611 
- This draft PR which identifies some strangeness in what methods are precompiled: https://github.com/JuliaLang/julia/pull/54155
- This draft PR (which was the starting point for this PR, thanks @vchuravy) https://github.com/JuliaGPU/CUDA.jl/pull/2336 which uses `@device_function` to provide erroring copies of the methods for the global method table

### DIFF
--- a/test/core/device/ldg.jl
+++ b/test/core/device/ldg.jl
@@ -1,5 +1,9 @@
 @testset "ldg" begin
-    ir = sprint(io->CUDA.code_llvm(io, CUDACore.pointerref_ldg, Tuple{Core.LLVMPtr{Int,AS.Global},Int,Val{1}}; raw=true))
+    # NOTE: This is necessary because it seems that code_llvm has a bug which causes it to ignore
+    #       the method table. Wrapping it in a function gets us what we want currently but the PR
+    #       here: https://github.com/JuliaLang/julia/pull/60718 will likely fix this according to
+    #       @vchuravy. It is currently not backported.
+    ir = sprint(io->CUDA.code_llvm(io, (args...)->CUDA.pointerref_ldg(args...), Tuple{Core.LLVMPtr{Int,AS.Global},Int,Val{1}}; raw=true))
     if Base.libllvm_version >= v"20"
         # `@llvm.nvvm.ldg` was removed in LLVM 20; the auto-upgrade
         # replaces it with a load bearing `!invariant.load` metadata
--- a/test/core/device/ldg.jl
+++ b/test/core/device/ldg.jl
@@ -3,7 +3,7 @@
     #       the method table. Wrapping it in a function gets us what we want currently but the PR
     #       here: https://github.com/JuliaLang/julia/pull/60718 will likely fix this according to
     #       @vchuravy. It is currently not backported.
-    ir = sprint(io->CUDA.code_llvm(io, (args...)->CUDA.pointerref_ldg(args...), Tuple{Core.LLVMPtr{Int,AS.Global},Int,Val{1}}; raw=true))
+    ir = sprint(io->CUDA.code_llvm(io, (args...)->CUDACore.pointerref_ldg(args...), Tuple{Core.LLVMPtr{Int,AS.Global},Int,Val{1}}; raw=true))
     if Base.libllvm_version >= v"20"
         # `@llvm.nvvm.ldg` was removed in LLVM 20; the auto-upgrade
         # replaces it with a load bearing `!invariant.load` metadata
--- a/CUDACore/src/device/intrinsics/atomics.jl
+++ b/CUDACore/src/device/intrinsics/atomics.jl
@@ -151,7 +151,7 @@ for A in (AS.Generic, AS.Global, AS.Shared), T in (:Int16, :UInt16)
     end
 
     intr = "atom$scope.cas.b16 \$0, [\$1], \$2, \$3;"
-    @eval @inline atomic_cas!(ptr::LLVMPtr{$T,$A}, cmp::$T, val::$T) =
+    @eval @device_function @inline atomic_cas!(ptr::LLVMPtr{$T,$A}, cmp::$T, val::$T) =
         @asmcall($intr, "=h,l,h,h", true, $T, Tuple{Core.LLVMPtr{$T,$A},$T,$T}, ptr, cmp, val)
 end
 
@@ -172,7 +172,7 @@ for A in (AS.Generic, AS.Global, AS.Shared)
         nb = sizeof(T)*8
         fn = Symbol("atomic_$(op)!")
         intr = "llvm.nvvm.atomic.load.$op.$nb.p$(convert(Int, A))i$nb"
-        @eval @inline $fn(ptr::LLVMPtr{$T,$A}, val::$T) =
+        @eval @device_function @inline $fn(ptr::LLVMPtr{$T,$A}, val::$T) =
             @typed_ccall($intr, llvmcall, $T, (LLVMPtr{$T,$A}, $T), ptr, val)
     end
 end
@@ -192,7 +192,7 @@ for A in (AS.Generic, AS.Global, AS.Shared), T in (:Float16,)
     end
 
     intr = "atom$scope.add.noftz.f16 \$0, [\$1], \$2;"
-    @eval @inline atomic_add!(ptr::LLVMPtr{$T,$A}, val::$T) =
+    @eval @device_function @inline atomic_add!(ptr::LLVMPtr{$T,$A}, val::$T) =
         @asmcall($intr, "=h,l,h", true, $T, Tuple{Core.LLVMPtr{$T,$A},$T}, ptr, val)
 end
 
--- a/CUDACore/src/device/intrinsics/cooperative_groups.jl
+++ b/CUDACore/src/device/intrinsics/cooperative_groups.jl
@@ -561,7 +561,8 @@ end
 
 ## pipeline operations
 
-@device_function pipeline_commit() = ccall("llvm.nvvm.cp.async.commit.group", llvmcall, Cvoid, ())
+@device_function pipeline_commit() =
+    ccall("llvm.nvvm.cp.async.commit.group", llvmcall, Cvoid, ())
 
 @device_function pipeline_wait_prior(n) =
     ccall("llvm.nvvm.cp.async.wait.group", llvmcall, Cvoid, (Int32,), n)
--- a/CUDACore/src/device/intrinsics/cooperative_groups.jl
+++ b/CUDACore/src/device/intrinsics/cooperative_groups.jl
@@ -24,7 +24,7 @@ Noteworthy missing functionality:
 module CG
 
 using ..CUDACore
-using ..CUDACore: i32, Aligned, alignment
+using ..CUDACore: i32, Aligned, alignment, @device_function
 
 import ..LLVM
 using ..LLVM.Interop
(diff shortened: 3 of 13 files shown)


KSepetanc:
Can confirm. Error `Cannot select: intrinsic %llvm.nvvm` does not appear with this changes any more. Thank you!!

Remaining issue is as you state at point 4. Previously it was observed in [comment](https://github.com/JuliaGPU/GPUCompiler.jl/issues/611#issuecomment-3614294092) when trying to AOT compile GPUCompiler. 

I assume `GPUCompiler.jl/src/runtime.jl:85` and its fixme note need to be checked.

apozharski:
> I assume GPUCompiler.jl/src/runtime.jl:85 and its fixme note need to be checked.

I started some digging into this yesterday and I think this is a bit of a red herring. If this was the core issue I would expect the methods that leak through to be all of the ones that need to be specified by the target or only the ones that are provided defaults. The two that currently are causing issues are `signal_exception` and `gc_pool_alloc` (which contains calls to `malloc` and `report_oom`) the first of which is target provided and the latter has a default implementation. I suspect there is something a little more subtle going on. I have some more time today to look at this so hopefully I will have at least a draft fix for that part soon.

michel2323:
I've started to mark all the methods in CUDA.jl as device methods. I still have to clean it up. Using @apozharski 's branch https://github.com/apozharski/GPUCompiler.jl/tree/ap/aot-compilation I can now get past the compilation. However, the next hurdle is that `CUDA_Drive_jll.jl` tries run julia and I get this error from the `_init_()` here: https://github.com/JuliaBinaryWrappers/CUDA_Driver_jll.jl/blob/039474600cf989a464f195dc573326e8cfa6a6cc/src/wrappers/x86_64-linux-gnu.jl#L9

```julia
Core.InitError(mod=:CUDA_Driver_jll, error=Base.TaskFailedException(task=Core.Task(next=nothing, queue=nothing, storage=nothing, donenotify=Base.GenericCondition{Base.Threads.SpinLock}(waitq=Base.IntrusiveLinkedList{Core.Task}(head=nothing, tail=nothing), lock=Base.Threads.SpinLock(owned=0)), result=Base.IOError(msg="could not spawn setenv(`/home/michel/git/ExaPF.jl/app/build/bin/julia -C native -g1 -O0 --compile=min -t1 --startup-file=no -e '...'
```

apozharski:
> I've started to mark all the methods in CUDA.jl as device methods. I still have to clean it up. Using @apozharski 's branch https://github.com/apozharski/GPUCompiler.jl/tree/ap/aot-compilation I can now get past the compilation. However, the next hurdle is that `CUDA_Drive_jll.jl` tries run julia and I get this error from the `_init_()` here: https://github.com/JuliaBinaryWrappers/CUDA_Driver_jll.jl/blob/039474600cf989a464f195dc573326e8cfa6a6cc/src/wrappers/x86_64-linux-gnu.jl#L9
> 
> ```julia
> Core.InitError(mod=:CUDA_Driver_jll, error=Base.TaskFailedException(task=Core.Task(next=nothing, queue=nothing, storage=nothing, donenotify=Base.GenericCondition{Base.Threads.SpinLock}(waitq=Base.IntrusiveLinkedList{Core.Task}(head=nothing, tail=nothing), lock=Base.Threads.SpinLock(owned=0)), result=Base.IOError(msg="could not spawn setenv(`/home/michel/git/ExaPF.jl/app/build/bin/julia -C native -g1 -O0 --compile=min -t1 --startup-file=no -e '...'
> ```

@michel2323 the workaround I found for this while trying to compile [libMad](https://github.com/apozharski/libMad) is to bundle the julia executable (see the [latest commit on my fork of JuliaC](https://github.com/apozharski/JuliaC.jl/commit/b939c14671b9d0866cd305eac455e185bfdc58f8)). Ideally `CUDA_driver_jll` would not fork to another julia process as this is likely to be very annoying to get to be relocatable, but this seems to work.

michel2323:
Nice! Thank you! @apozharski

test/core/device/ldg.jl
```diff
@@ -1,8 +1,3 @@
-@testset "ldg" begin
-    ir = sprint(io->CUDA.code_llvm(io, CUDA.pointerref_ldg, Tuple{Core.LLVMPtr{Int,AS.Global},Int,Val{1}}))
-    @test occursin("@llvm.nvvm.ldg", ir)
-end
```
maleadt, line 10: Why remove this test? `CUDA.code_llvm` (as opposed to the `InteractiveUtils` version) should use the correct method table, so the generated code should be as expected.

maleadt:
> Ideally `CUDA_driver_jll` would not fork to another julia process as this is likely to be very annoying to get to be relocatable, but this seems to work.

We used to do `dlopen` + `dlclose`, but that was extremely unreliable, causing segfaults on some users' systems. The `fork` approach seems like the best we can do given the dynamism of the system (users can upgrade their driver, swap hardware, etc. all of which impact the driver JLL).

test/core/device/ldg.jl
apozharski: Ah rebasing removed the easy way for me to find the CI run before this commit, but this test was certainly failing. You are right though, it shouldn't break. I will take another look.

apozharski:
> > Ideally `CUDA_driver_jll` would not fork to another julia process as this is likely to be very annoying to get to be relocatable, but this seems to work.
> 
> We used to do `dlopen` + `dlclose`, but that was extremely unreliable, causing segfaults on some users' systems. The `fork` approach seems like the best we can do given the dynamism of the system (users can upgrade their driver, swap hardware, etc. all of which impact the driver JLL).

This is out of scope, but a user changing their drivers/hardware while code is running causing segfaults seems to me like "expected" behavior :sweat_smile:. Unless you mean this only runs once and not each time a new Julia process calls `using CUDA_driver_jll` though I am not sure I entirely understand how the current solution fixes that.

apozharski:
> We used to do `dlopen` + `dlclose`, but that was extremely unreliable, causing segfaults on some users' systems. The `fork` approach seems like the best we can do given the dynamism of the system (users can upgrade their driver, swap hardware, etc. all of which impact the driver JLL).

Doing some more digging my little hack of bundling the julia executable (in truth I think this is a _horrible_ hack) is insufficient to work around this. This just means the run fails and the aot compiled object just _always_ falls back to loading the system version of the driver.  To get the hack to work correctly I would further need to include a standard sysimage as well in the distribution of a library which is just yet more bloat I would hope to avoid.

@maleadt would you be willing to accept a PR/change to `CUDA_driver_jll` which can fall back to a `dlopen`/`dlclose` approach on failure (or perhaps as a `Preference`?) as this would be _far_ more portable for any case that isn't REPL use?

apozharski:
@michel2323 this is also relevant to your work (as I misled you with the initial hack :sweat_smile: )

https://github.com/apozharski/CUDA_Driver_jll.jl/tree/ap/in-memory-driver-check <- This branch of `CUDA_Driver_jll` checks for device drivers in memory, removing the need for trying to bundle the `julia` executable with a library using `CUDA`. It currently has a bunch of debug printing as well but I will clean it up (and try to understand if some checks with `dlopen(RTLD_NOLOAD)` are necessary) and submit a PR hopefully today or tomorrow.

KSepetanc:
`@device_functions` is fixed now for Julia 1.13. Can this be put forward?

apozharski:
> `@device_functions` is fixed now for Julia 1.13. Can this be put forward?

I am currently stuck 100% on some other topics so pushing this over the line fell back in my priorities. I should be able to try to push this over the line in the next couple of weeks. 

This is the branch I have been using in my work and it "works on my machine" but could probably use some replication.

test/core/device/ldg.jl
apozharski: I reinstated this test during a rebase on the latest master, and it does indeed fail.
@maleadt I am a bit out of my depth but doing some brief debugging it seems somehow `code_llvm` (and probably the rest of `reflections.jl`) is broken now if you try to apply them to methods hidden in the method table.

I am not sure I understand why because it seems like the GPUCompiler job should see the correct method_table overlay.

apozharski:
Ok I now suspect that CI will pass (except on nightly it seems), so this is ready for re-review.

apozharski:
Hmmm, I am not sure the failure on CUDA 13.2 is related It seems to be an OOM issue which it seems occurred on previous commits as well.

KSepetanc:
It is failing now. You dropped last commit in rebase?

apozharski:
> It is failing now. You dropped last commit in rebase?

Ah yea, good catch

test/core/device/ldg.jl
maleadt: That change has been mostly backported, actually, only 1.13 is missing I think. So let's go with the workaround for now.

maleadt (approved):

apozharski:
> Hi @apozharski — great work on this PR! I believe 3 cluster barrier intrinsics were missed.
> 
> In `lib/cudacore/src/device/intrinsics/synchronization.jl`, lines 75-77:
> 
> ```julia
> cluster_arrive() = ccall("llvm.nvvm.barrier.cluster.arrive", llvmcall, Cvoid, ())
> cluster_arrive_relaxed() = ccall("llvm.nvvm.barrier.cluster.arrive.relaxed", llvmcall, Cvoid, ())
> cluster_wait() = ccall("llvm.nvvm.barrier.cluster.wait", llvmcall, Cvoid, ())
> ```
> 
> These aren't wrapped in `@device_functions`, unlike all other intrinsics in the same file. This causes `PackageCompiler.create_library()` to SIGABRT with:
> 
> ```
> LLVM ERROR: Cannot select: intrinsic %llvm.nvvm.barrier.cluster.wait
> ```
> 
> After wrapping them in `@device_functions begin ... end` (combined with [GPUCompiler.jl#808](https://github.com/JuliaGPU/GPUCompiler.jl/pull/808) for the `gpu_*` weak-linkage stubs), full AOT compilation succeeds.
> 
> Tested with Julia 1.12.6, CUDA.jl 6.1.0, CUDACore 6.1.1, GPUCompiler 1.13.2 (from `main`).

`lib/cudacore/src/device/intrinsics/synchronization.jl` does not exist. You mean `CUDACore/src/device/intrinsics/synchronization.jl`? Your em-dash loving stochastic parrot is hallucinating again.

However,  these were missed in the rebase, though somehow this issue does not prevent `libMad` from compiling. Why you feel the need to tag me in a closed PR is beyond me, but I suspect that I am trying to ascribe reason to an unthinking machine.

PR #3003, opened 2025-12-24
[CI] Merge docker images across platforms

giordano:
With the current setup, images for different platforms are pushed with the same tag, overriding each other.  With this change, instead, images built for different architectures are merged together.  I still need to do some more testing, but something like this should do the trick.

### DIFF
--- a/.github/workflows/Container.yml
+++ b/.github/workflows/Container.yml
@@ -1,3 +1,7 @@
+# Based on:
+# * <https://docs.github.com/en/actions/publishing-packages/publishing-docker-images>
+# * <https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners>
+
 name: Publish Docker container
 
 on:
@@ -18,8 +22,12 @@ on:
     branches:
       - main
 
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
 jobs:
-  push_to_registry:
+  build:
     name: Container for ${{ matrix.platform }} - Julia ${{ matrix.julia }} - CUDA ${{ matrix.cuda }}
     permissions:
       contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/Container.yml
```diff
@@ -90,37 +109,99 @@ jobs:
...
-        uses: docker/metadata-action@v5
-        with:
-          images: ghcr.io/${{ github.repository }}
-          tags: |
-              type=raw,value=${{ steps.pkg.outputs.name }}-julia${{ matrix.julia }}-cuda${{ steps.cuda.outputs.major }}
-              type=raw,value=${{ steps.pkg.outputs.name }},enable=${{ matrix.default == true && (github.ref_type == 'tag' || inputs.tag != '') }}
-              type=raw,value=latest,enable=${{ matrix.default == true && (github.ref_type == 'tag' || (inputs.tag != '' && inputs.mark_as_latest)) }}
-              type=raw,value=dev,enable=${{ matrix.default == true && github.ref_type == 'branch' && inputs.tag == '' }}
-          labels: |
-              org.opencontainers.image.version=${{ steps.pkg.outputs.version }}
```
giordano: I honestly don't quite understand what was going on here.  If anything needs to be adapted, it should be done in the `tags` input in the `Build and push image` step below.

giordano:
See for example https://github.com/giordano/CUDA.jl/pkgs/container/cuda.jl/622086517?tag=dev-julia1.11-cuda12, where the tag has both architectures, instead of only the last one which happened to have been pushed.

.github/workflows/Container.yml
maleadt: When forcing the workflow, there's params that can be set: 

Image

The idea is to extract that here, and set that as metadata on the generated containers so that you can do `docker run ghcr.io/juliagpu/cuda.jl:dev` or `ghcr.io/juliagpu/cuda.jl:cuda-12` etc.

PR #3007, opened 2026-01-02
Add native BFloat16 `atomic_add!` support

AntonOresten:
Note that sequentially accumulating ones in BFloat16 precision will fail after 256, so the tests use fewer threads for BFloat16.

The runners don't seem to have the required CC of 9.0 to hit the tests.

### DIFF
--- a/src/device/intrinsics/atomics.jl
+++ b/src/device/intrinsics/atomics.jl
@@ -180,9 +180,9 @@ end
 
 ## PTX
 
-# half-precision atomics using PTX instruction
+# half-precision and bfloat16 atomics using PTX instructions
 
-for A in (AS.Generic, AS.Global, AS.Shared), T in (:Float16,)
+for A in (AS.Generic, AS.Global, AS.Shared), T in (:Float16, :BFloat16)
     if A == AS.Global
         scope = ".global"
     elseif A == AS.Shared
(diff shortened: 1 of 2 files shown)


AntonOresten:
I ran the example from the first comment again on master and am confused by it working. Something may have been wrong with my setup.

EDIT: I had to backtrack to understand what was happening. `CUDA.@atomic` seems to have worked with BFloat16 even before, but not `KA.@atomic`, which uses some `modify!` from Atomix.jl, so BFloat16 wasn't hooking up to the necessary interfaces:

```julia
julia> Zygote.gradient(x .|> BFloat16) do x
           ONIONop.flash_attention(x, x, x; causal=true) |> sum
       end[1]
ERROR: InvalidIRError: compiling MethodInstance for ONIONop.gpu__flash_attention_bwd!(::KernelAbstractions.CompilerMetadata{…}, ::Type{…}, ::Type{…}, ::Type{…}, ::Type{…}, ::Type{…}, ::CuDeviceArray{…}, ::CuDeviceArray{…}, ::CuDeviceArray{…}, ::CuDeviceArray{…}, ::CuDeviceArray{…}, ::CuDeviceArray{…}, ::CuDeviceArray{…}, ::CuDeviceArray{…}, ::CuDeviceArray{…}, ::CuDeviceArray{…}, ::CuDeviceArray{…}, ::BFloat16, ::Nothing, ::Nothing, ::Val{…}, ::Val{…}, ::Val{…}, ::Val{…}) resulted in invalid LLVM IR
Reason: unsupported call to an unknown function (call to jl_f_throw_methoderror)
Stacktrace:
 [1] modify!
   @ ~/.julia/packages/Atomix/0UMek/ext/AtomixCUDAExt.jl:38
 [2] macro expansion
   @ ~/.julia/packages/KernelAbstractions/X5fk1/src/extras/loopinfo.jl:31
 [3] macro expansion
   @ ~/.julia/packages/ONIONop/m70QD/src/attention/attention_bwd.jl:103
 [4] gpu__flash_attention_bwd!
   @ ./none:0
```

In the midst of having a partially complete implementation, I must've forgotten to remove a line when comparing.

PR #3017, opened 2026-01-13
Add support for thread block clusters

eschnett:


lib/cudadrv/libcuda.jl
```diff
@@ -2258,15 +2258,15 @@ function Base.getproperty(x::Ptr{CUlaunchAttributeValue_union}, f::Symbol)
     f === :accessPolicyWindow && return Ptr{CUaccessPolicyWindow}(x + 0)
     f === :cooperative && return Ptr{Cint}(x + 0)
     f === :syncPolicy && return Ptr{CUsynchronizationPolicy}(x + 0)
-    f === :clusterDim && return Ptr{var"##Ctag#247"}(x + 0)
+    f === :clusterDim && return Ptr{CuDim3}(x + 0)
```
maleadt: These are auto-generated, so will get lost.

lib/cudadrv/execution.jl
```diff
@@ -44,22 +44,29 @@ end
...
 function launch(f::CuFunction, args::Vararg{Any,N}; blocks::CuDim=1, threads::CuDim=1,
-                cooperative::Bool=false, shmem::Integer=0,
+                clustersize::CuDim=1, cooperative::Bool=false, shmem::Integer=0,
                 stream::CuStream=stream()) where {N}
     blockdim = CuDim3(blocks)
     threaddim = CuDim3(threads)
+    clusterdim = CuDim3(clustersize)
+    if CUDA.capability(device()) < v"9.0" && (clusterdim.x != 1 || clusterdim.y != 1 || clusterdim.y != 1)
+        throw(ArgumentError("devices with compute capability less than 9.0 do not support thread block clusters!"))
+    end
```
maleadt, line 82: There's other validation to be done:
> Note that it is illegal for the second tuple of __block_size__ and __cluster_dims__ to be specified at the same time. It’s also illegal to use __block_size__ with an empty __cluster_dims__. When the second tuple of __block_size__ is specified, it implies the “Blocks as Clusters” being enabled and the compiler would recognize the first argument inside <<<>>> as the number of clusters instead of thread blocks.
Probably something for the launch failure diagnosis code.

maleadt: I'm not sure we want to do a `capability` API call during launch. What about diagnosing this post-hoc instead?

maleadt (changes requested):

lib/cudadrv/libcuda.jl
eschnett: Is there a way to tell the generator to convert `Dim3` to our `CuDim3`?

lib/cudadrv/execution.jl
eschnett: We're not supporting the syntax where `block_size` consists of two tuples. We're using a lower-level approach where the block size and grid size are passed as launch arguments, and the cluster size as launch attribute. I don't see what additional validations we'd need to do since we're mapping to the CUDA API in a straightforward way.

I can drop the capability check.

lib/cudadrv/libcuda.jl
eschnett: There isn't. I reverted these changes.

eschnett:
For the record, this PR addresses https://github.com/JuliaGPU/CUDA.jl/issues/1989.

eschnett:
I'll have to clean up the patch, but apart from that it's finished.

- There is a way to launch kernels with non-trivial cluster sizes
- There are new kernel functions to query about clusters
- There is distributed shared memory
- There is a `cluster_wait` function.

Missing cleanups:
- Probably needs more `device_capability >= v"9.0"` checks
- Requires LLVM >= 20, i.e. Julia >= 1.13; will need to add `@static if` statements for this
- Need to make tests pass
- CUDA.jl is broken for Julia 1.13. I have a work-around, but I assume someone else will clean this up, and I'll then have to merge it

lib/cudadrv/execution.jl
```diff
@@ -68,29 +72,46 @@ function launch(f::CuFunction, args::Vararg{Any,N}; blocks::CuDim=1, threads::Cu
...
+                GC.@preserve attr_ref stream begin
+                    attr = Base.unsafe_convert(Ptr{CUDA.CUlaunchAttribute}, attr_ref)
+                    attr.id = CUDA.CU_LAUNCH_ATTRIBUTE_CLUSTER_DIMENSION
+                    attr.value.clusterDim.x = clusterdim.x
+                    attr.value.clusterDim.y = clusterdim.y
+                    attr.value.clusterDim.z = clusterdim.z
+                    config = Ref(CUlaunchConfig(blockdim.x, blockdim.y, blockdim.z,
+                                                threaddim.x, threaddim.y, threaddim.z,
+                                                shmem, stream.handle, attr, 1))
+                    cuLaunchKernelEx(config, f, kernelParams, C_NULL)
```
maleadt: We can probably switch to `cuLaunchKernelEx` for all code paths? We require CUDA 12.0, after all, which AFAIU is the version that introduced `cuLaunchKernelEx`.

lib/cudadrv/execution.jl
```diff
@@ -68,29 +72,46 @@ function launch(f::CuFunction, args::Vararg{Any,N}; blocks::CuDim=1, threads::Cu
...
             end
         end
     catch err
-        diagnose_launch_failure(f, err; blockdim, threaddim, shmem)
+        diagnose_launch_failure(f, err; blockdim, threaddim, clusterdim, shmem)
     end
 end
 
-@noinline function diagnose_launch_failure(f::CuFunction, err; blockdim, threaddim, shmem)
+@noinline function diagnose_launch_failure(f::CuFunction, err; blockdim, threaddim, clusterdim, shmem)
```
maleadt: Maybe add a compute capability check? clusterdim should be 1 on pre-sm_90.

maleadt (commented):
LGTM. Couple of minor nits.

lib/cudadrv/execution.jl
eschnett: Done.

lib/cudadrv/execution.jl
eschnett: Done.

eschnett:
@maleadt I removed the nits.

eschnett:
Could you merge this PR?

PR #3026, opened 2026-02-03
Modify range metadata in indexing.jl

### LINKED ISSUE #3024: Indexing range metadata is set up wrong
The file `indexing.jl` contains the lines
```
            # attach range metadata
            range_metadata = MDNode([ConstantInt(Int32(range.start)),
                                     ConstantInt(Int32(range.stop))])
            metadata(idx)[LLVM.MD_range] = range_metadata
```

LLVM range metadata https://llvm.org/docs/LangRef.html#range-metadata are described by a half-open interval, as is common in C++. Thus `range.stop` needs to be `range.stop + 1` in the code above.

This error leads to weird code generation problems when checking block or grid sizes just below the maximum possible.

(Having the lower bound there is a good idea. I am not sure about the upper bounds as they may increase in newer architectures, and without updating the bounds here we would generate wrong code for these architectures.)

SuyashR25:
Fixes #3024 

Update range metadata to include stop + 1.


SuyashR25 (commented):
Issues #3024 
Fixed metadata range stop to stop + 1

maleadt (approved):

maleadt (changes requested):
This seems to break things spectacularly, throwing an error from the `_index` generated function.

maleadt:
Replaced by https://github.com/JuliaGPU/CUDA.jl/pull/3119

Issue #2859, opened 2025-08-27
Any strong reason to have DataFrames.jl as a dependency?


juliohm:
It is a heavy dependency, and I wonder if it could be replaced by built-in table types (e.g., named tuples of vectors) provided by the language.

kshyatt:
It's there for the profiler

juliohm:
I'll try to submit a PR to remove the dependency in favor of native Julia
tables. A named tuple of vectors should do it, specially if all that is
needed is a pretty_table call.

Em qua., 27 de ago. de 2025, 13:05, Katharine Hyatt <
***@***.***> escreveu:

juliohm:
Unfortunately the code relies on `leftjoin` and other data access patterns that are specific to DataFrames.jl.

I don't have the time to fix it myself at the moment, but it would be great if this dependency could be dropped in the future. Ideally, the profiler would be written in terms of the Tables.jl API with built-in types (e.g., named tuple of vectors, vector of named tuples).

GunnarFarneback:
Cf. earlier discussion in https://github.com/JuliaGPU/CUDA.jl/issues/2238#issuecomment-2067984886.

maleadt:
Closing as a dup of https://github.com/JuliaGPU/CUDA.jl/issues/2238. I'm happy to take any PRs making the dependencies more lightweight, but as I mentioned I am currently of the belief that `using CUDA` should provide an environment that's similar to CUDA itself, which does include a bunch of tools like a profiler.

juliohm:
The profiler makes total sense. What doesn't make much sense is DataFrames.jl as a dependency to write this profiler in CUDA.jl. The DataFrames.jl package is at a higher-level of abstraction, and could potentially need CUDA.jl as an extension one day to implement GPU operations.

maleadt:
I'm using DataFrames.jl to post-process the CUPTI data, not to provide GPU integration with DataFrames.jl itself (which a package extension would be well suited for).

juliohm:
I know. What people are trying to say is that you can achieve the same result without adding DataFrames.jl as a dependency. That is why Tables.jl exists as a lightweight interface that works with built-in table types (named tuple of vectors, vector of named tuples).

Notice that CUDA.jl is currently locking updates on PrettyTables.jl v3 because of the DataFrames.jl dependency. It creates more friction than it helps.

maleadt:
[...]

PR #3029, opened 2026-02-08
Replace DataFrames in the profiler with NamedTuples

JamesWrigley:
As suggested in https://github.com/JuliaGPU/CUDA.jl/issues/2859 and https://github.com/JuliaGPU/CUDA.jl/issues/2238.

I initially tried to implement this using a Tables-esque API (`leftjoin()` etc) but it ended up being much simpler to implement that logic manually. One question I have is about the way that the `details` and `nvtx_data_lookup` dicts are used: they currently assume that the `record.correlationId` and `record.id` are unique, is that a safe assumption? I'm not too familiar with the profiler.

Load times:
```bash
# Before
$ hyperfine -w 2 -r 10 "julia --startup-file=no --project -e 'import CUDA'"
Benchmark 1: julia --startup-file=no --project -e 'import CUDA'
  Time (mean ± σ):      4.551 s ±  0.693 s    [User: 5.324 s, System: 0.366 s]
  Range (min … max):    4.180 s …  5.902 s    10 runs

# After
$ hyperfine -w 2 -r 10 "julia --startup-file=no --project -e 'import CUDA'"
Benchmark 1: julia --startup-file=no --project -e 'import CUDA'
  Time (mean ± σ):      3.702 s ±  0.684 s    [User: 4.488 s, System: 0.342 s]
  Range (min … max):    3.357 s …  5.000 s    10 runs
```

BTW when testing it I noticed that the `@bprofile` tests took like ~10 minutes on my node, which turned out to be the repeated calls to demumble. I modified the profiler in 7af4e5c42 to batch-call demumble and that removed most of the overhead, now `core/profile.jl` takes ~1 minute in total for me.

Mostly written by Claude, with various tweaks by me :robot:


JamesWrigley:
Not sure what's up with the 1.12 failures :thinking: It doesn't look related AFAICT.

giordano:
Same error happened also in https://github.com/JuliaGPU/CUDA.jl/commit/aa310ac70688f23b8273a2132c8b8f46cb23c443 on `master`.  I restarted the job 🤞

maleadt (approved):
LGTM, thanks!

PR #3030, opened 2026-02-13
Various improvements to the docs

giordano:
I had some...uhm...fun in the last couple of days trying to port some C++ CUDA code to CUDA.jl, and profile it.  I dumped into this PR my experience, hoping to make lives of people after me a little bit easier :slightly_smiling_face:


src/device/intrinsics/indexing.jl
```diff
@@ -63,54 +63,60 @@ end
...
-@doc """
-    threadIdx()::NamedTuple
+"""
+    gridDim()::NamedTuple
 
-Returns the thread index within the block.
-""" threadIdx
-@inline threadIdx() = (x=threadIdx_x(), y=threadIdx_y(), z=threadIdx_z())
+Returns the dimensions of the grid as a `NamedTuple` with keys `x`, `y`, and `z`.
+These dimensions have the same starting index as the `gridDim` built-in variable in the C/C++ extension.
```
maleadt: `gridDim` returns a dimension/size, not an index.

giordano: Replaced "index" with "dimension" here.

maleadt: `starting dimension` doesn't make much sense to me. What else could a `size()` query return? 0 vs 1-based indexing doesn't apply here.

That said, I'm okay with this if you think this clarifies things.

christiangnrd: Maybe it could be phrased along the lines of:
```
Unlike the `*Idx` intrinsics `gridDim` returns the same value as its C/C++ extension counterpart.
```

I do think this should be mentioned in form though. The indexing intrinsics being offset while the dim intrinsics not makes sense when you think about it, but I've also gotten confused by this, and not everyone will think/know to check the source code to confirm.

Either way, the same edits the `gridDim` receives should also be mirrored to `blockDim`

giordano: Done.

giordano:
Bump?  I expanded also doscstrings introduced in #3017.

giordano:
Bump.  This PR keeps running into conflicts with other PRs which are merged in the meantime... 🫠

maleadt:
Sorry, forgot about this. Thanks!

PR #3032, opened 2026-02-20
[CUSPARSE] create slices of sparse matrices using boolean masks

hexaeder:
Slicing into a sparse matrix by providing `Vector{Bool}` per axis to select a subset of rows and cols is quite usefull. It is implemented for several types, for example "normal" matrices and sparse matrices.

```juila
N = 10
A = sprand(N, N, 0.5)
rowmask = [rand([true, false]) for _ in 1:size(A, 1)]
colmask = [rand([true, false]) for _ in 1:size(A, 2)]
S = A[rowmask, colmask]
```
This PR implements this functionality for `CuSparseMatrixCSR` and `CuSparseMatrixCSC`.

AI Disclaimer: Claude code and Chat GPT helped me to implement the kernels. But I think they are straight forward and it is tested against the slicing results of CPU code.

### DIFF
--- a/lib/cusparse/src/array.jl
+++ b/lib/cusparse/src/array.jl
@@ -522,52 +522,89 @@ function Base.getindex(A::CuSparseArrayCSR{Tv, Ti, N}, i0::Integer, i1::Integer,
 end
 
 # slice matrix by masking rows and columns
-function _getindex_boolmask(A::Union{CuSparseMatrixCSR, CuSparseMatrixCSC},
-                            Imask::CuVector{Bool}, Jmask::CuVector{Bool})
+
+function Base.getindex(A::CuSparseMatrixCSR{Tv,Ti}, Imask::CuVector{Bool}, Jmask::CuVector{Bool}) where {Tv,Ti}
     m, n = size(A)
     length(Imask) == m || throw(DimensionMismatch("boolean row mask length $(length(Imask)) must match row count $m"))
     length(Jmask) == n || throw(DimensionMismatch("boolean column mask length $(length(Jmask)) must match column count $n"))
 
-    # convert to COO for uniform element-wise access
-    A_coo = CuSparseMatrixCOO(A)
-    rowInd = A_coo.rowInd
-    colInd = A_coo.colInd
-    nzVal  = A_coo.nzVal
+    rowmap = cumsum(Ti.(Imask))
+    colmap = cumsum(Ti.(Jmask))
+    new_m = Int(CUDA.@allowscalar rowmap[end])
+    new_n = Int(CUDA.@allowscalar colmap[end])
 
-    # build selection mask: keep entry if both row and col are selected
-    row_sel = Imask[rowInd]
-    col_sel = Jmask[colInd]
-    keep = row_sel .& col_sel
+    # pass 1: count kept entries per new row
+    counts = CUDA.zeros(Ti, new_m)
+    if new_m > 0 && new_n > 0
+        threads = min(256, m)
+        blocks = cld(m, threads)
+        @cuda threads=threads blocks=blocks _csr_count_kernel!(counts, A.rowPtr, A.colVal, Imask, Jmask, rowmap)
+    end
 
-    # filter entries
-    new_rowInd = rowInd[keep]
-    new_colInd = colInd[keep]
-    new_nzVal  = nzVal[keep]
+    # build new rowPtr from counts: [1, 1+cumsum(counts)...]
+    new_rowPtr = vcat(CuVector{Ti}([one(Ti)]), cumsum(counts) .+ one(Ti))
+    new_nnz = Int(CUDA.@allowscalar new_rowPtr[end]) - 1
 
-    # build index remapping via cumulative sum of masks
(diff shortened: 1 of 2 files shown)


hexaeder:
The new tests uncovered a CSC to CSR conversion error in CUDA runtime 12.0 (fixed in 12.1 and above).
The `colVal` stored in the CSR matrix is just wrong, which messes printout, indexing and everything.
Should this be tracked or is it not usefull since it appears to be an upstream bug of that CUDA version?

I updated the tests to go through the COO format which hopefully avoids the buggy codepath.

MWE on Runtime 12.0
```julia
# version 12.0
julia> using CUDA, SparseArrays, CUDA.CUSPARSE

julia> A = sparse([1,1,1,1,2], [1,2,4,5,3], [1.0,2.0,3.0,4.0,5.0], 2, 5)
2×5 SparseMatrixCSC{Float64, Int64} with 5 stored entries:
 1.0  2.0   ⋅   3.0  4.0
  ⋅    ⋅   5.0   ⋅    ⋅

julia> dA = CuSparseMatrixCSR(A) # wrong!
2×5 CuSparseMatrixCSR{Float64, Int32} with 5 stored entries:
 1.0  2.0  5.0  3.0   ⋅
  ⋅    ⋅    ⋅    ⋅   4.0

julia> CUDA.@allowscalar dA[1,1] # wrong!
0.0
```
MWE on Runtime 12.1
```julia
# version 12.1
julia> using CUDA, SparseArrays, CUDA.CUSPARSE

julia> A = sparse([1,1,1,1,2], [1,2,4,5,3], [1.0,2.0,3.0,4.0,5.0], 2, 5)
2×5 SparseMatrixCSC{Float64, Int64} with 5 stored entries:
 1.0  2.0   ⋅   3.0  4.0
  ⋅    ⋅   5.0   ⋅    ⋅

julia> dA = CuSparseMatrixCSR(A) # correct!
2×5 CuSparseMatrixCSR{Float64, Int32} with 5 stored entries:
 1.0  2.0   ⋅   3.0  4.0
  ⋅    ⋅   5.0   ⋅    ⋅

julia> CUDA.@allowscalar dA[1,1] # correct!
1.0
```

hexaeder:
Gentle bump on this. Is there something I can do to support review?

EDIT: I rebased, github UI was showing some apparent merge conflicts without showing my where the problems lie.

kshyatt:
Looks like this might need some TLC to get up to date with the (unreleased) CUDA.jl 6 and package splitting

hexaeder:
I think the problems related to the Library split are fixed (just some CUDA -> CUDACore renamings). There are three remaning test fails I don't quite understand but seem unrelated?
- [Multi GPU Tests](https://buildkite.com/julialang/cuda-dot-jl/builds/7109/steps/canvas?sid=019d6715-2873-4a3b-9257-22b48af598cf) failed on `core/array` tests
- [Julia Nightly](https://buildkite.com/julialang/cuda-dot-jl/builds/7109/steps/canvas?jid=019d6715-28cb-4f5a-9ccd-bff07050fd66) fails because of out of bounds indexing in `core/array` test
- [Enzyme tests](https://buildkite.com/julialang/cuda-dot-jl/builds/7109/steps/canvas?sid=019d6715-286b-4c02-b348-65ba8b4db465) fail

maleadt:
Nightly and Enzyme are allowed to fail, and the multiGPU one is intermittent. So no relevant CI failures.

lib/cusparse/src/array.jl
```diff
@@ -521,6 +521,94 @@ function Base.getindex(A::CuSparseArrayCSR{Tv, Ti, N}, i0::Integer, i1::Integer,
...
+
+function Base.getindex(A::CuSparseMatrixCSR{Tv, Ti}, Imask::CuVector{Bool}, Jmask::CuVector{Bool}) where {Tv, Ti}
+    m, n = size(A)
+    length(Imask) == m || throw(DimensionMismatch("boolean row mask length $(length(Imask)) must match row count $m"))
+    length(Jmask) == n || throw(DimensionMismatch("boolean column mask length $(length(Jmask)) must match column count $n"))
+
+    rowmap = cumsum(Ti.(Imask))
+    colmap = cumsum(Ti.(Jmask))
+    new_m = Int(CUDACore.@allowscalar rowmap[end])
+    new_n = Int(CUDACore.@allowscalar colmap[end])
```
maleadt: What if masks are empty?

```julia-repl
julia> A = CuSparseMatrixCSR(spzeros(Float32, 0, 5))
0×5 CuSparseMatrixCSR{Float32, Int32} with 0 stored entries

julia> A[CuVector{Bool}([]), CuVector(rand(Bool, 5))]
ERROR: BoundsError: attempt to access 0-element CuArray{Int64, 1, CUDACore.DeviceMemory} at index [0]
```

hexaeder: Thanks for the hint, I addressed the problem by adding length=0 guards and more tests for corner cases :)

maleadt (approved):
LGTM. Too bad about the scalar `getindex`es that essentially synchronize the GPU before doing any work, but I'm not familiar with the domain to suggest alternatives here.

PR #3040, opened 2026-03-04
Initial support for CUDA 13.1.

maleadt:


kshyatt:
Is the plan to handle actually wrapping some of the CUBLAS emulation stuff separately?

maleadt:
Not on my end. But looking at the API, it shouldn't be too hard; just another flag to be configured on the handle.

PR #3050, opened 2026-03-07
Move nvtx initialization to first profile

wsmoses:
cc @giordano @avik-pal [not necessarily a solution to our problems at the moment given our previous workaround, but this should regardless hopefully stabilize things a bit]

### DIFF
--- a/src/initialization.jl
+++ b/src/initialization.jl
@@ -150,12 +150,6 @@ function __init__()
         end
     end
 
-    # if we're not running under an external profiler, let CUPTI handle NVTX events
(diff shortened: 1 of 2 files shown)


maleadt (changes requested):
This may break profiling entire applications with NVTX metadata when not calling `CUDA.@profile` (which is not required when doing entire-application profiling). Let me verify.

maleadt (approved):
Verified that `nsys` sets the env var itself, so that use case remains functional. Note that this now also depends on NVTX.jl's `__init__` calling `activate()` when `NVTX_INJECTION64_PATH` is detected.

PR #3052, opened 2026-03-11
FFT plans reworked

### LINKED ISSUE #119: CUFFT: support for arbitrary dims
A 1d fft across the nth dimension of > n dimensional CuArray (where n is not 1) is not enabled by the wrapper  (ERROR: ArgumentError: batching dims must be sequential    from CuArrays>wXQp8>src>fft>wrappers.jl) 

to reproduce: 
using CuArrays, CUDAnative, CUDAdrv, FFTW
dim = 2
data = CuArrays.rand(ComplexF32,512,512,512);
myFft = plan_fft!(data,dim);

the wrapper for regular cpu arrays allows this, and if dim is 1 or 3 it also works as expected

RainerHeintzmann:
In a use-case a problem with FFTs showed up, that previously surfaced a number of times. This lead to a reworking of the FFT implementations. Here is the example that caused massive problems:
```julia
julia> using CUDA, FFTW
julia> d = rand(Float32, 7,7,1000000);
julia> dc = cu(d);
julia> CUDA.@time dcf = fft(dc, 1);
julia> CUDA.@time dcf = fft(dc, 2);
```
The latter command (y-direction FFT) is in the previous code insanely slow (~15 seconds) as opposed to 0.1 sec for the X direction.
The reason is the for loop over the "trailing" batch dimensions, and the 1 million kernel starts, which is 100x slower than internal batching. I tried some approaches using streams but with no success, which led to rewriting the planning algorithm.
I find it quite essential to have a decently fast Y-transform, since in many implementations of ND-FFTs or (in our case the CZT algorithm in `FourierToosl.jl`) transforms are consecutively applied in the various directions.

In addition, there previously were unnecessary restrictions to possible combinations of transform directions and the dimensions always had to be order from low to high.
In the current version:
+ The code is shortened, a number of if-branches have been removed, comments were added and thus it is hopefully now easier to understand
+ The previous division between leading=internal and trailing = external batch dimensions has been removed in favor of a speed-related choice.
+ Is now dynamically determined which is the best combination of consecutive "internal" batch dimensions, and all other non-transform directions are now treated as "external". Therefore these is no more restriction to combinations of transform directions except:  1) They still have to be three or less transform directions and 2) for ComplexF16 and Float16 there are some alignment restrictions imposed by the CuFFT framework. 
+ For FFTs the dims are automatically sorted before the transform, so the input can be in any order. However, this is intentionally not the case for rfft, since in the FFTW library this has the meaning of halfing the size of the first mentioned dimension, not the lowest one. Since this behaviour is sitll NOT supported, an error is thrown for non rising order number.
+ A slight drawback of the new was of batching is, that rfft of any `Real` input type, in which the external batch dimensions lead to an `idist` stride of uneven size, require the application of two in-place `circshift!` on the input array between batching over each second element. Maybe this problem, which throws an alignment error, existed even previously but never showed up or was never reported.

P.S.: Sorry for having updated the dependencies without proper testing of the whole `CUDA.jl`. You can simply revert the changes to `Project.toml`

Fixes #119


RainerHeintzmann:
@maleadt, is there a way to make it pass the buildkite/cud-dot-jl ?
The changes only concerned the lib/CuFFT stuff and they pass the local tests.

christiangnrd:
Failures seem related. Sorting NTuples from Base is only available as of Julia 1.12

RainerHeintzmann:
> Failures seem related. Sorting NTuples from Base is only available as of Julia 1.12

Thanks! Great catch. So I will fix this and push again.

RainerHeintzmann:
> Failures seem related. Sorting NTuples from Base is only available as of Julia 1.12

That seems to have worked but other packages (sparse) seem to fail using the CUDA 13.2 configuration.

RainerHeintzmann:
@maleadt, @christiangnrd   Thanks for updating `CUDA.jl` such that all checks successfully pass now. What do you think about merging this PR with the main branch, such that FFTs get a boost?

maleadt:
Rebased. I'm not sure how to review this, though, as I'm really not familiar with the inner working of FFT. I tried to do some LLM-assisted review, and came up with the following issues that look real:

> ### Bug 1: `get_batch_dims` crashes on 5D+ arrays with non-contiguous transform dims
> 
> ```julia
> julia> fft(CUDA.rand(ComplexF32, 3, 5, 7, 11, 2), (2, 4))
> ERROR: MethodError: no method matching Tuple(::Int64, ::UnitRange{Int64})
> # at fft.jl:163
> ```
> 
> The line `external_batch_dims = Tuple(external_batch_dims..., (previous_transform_dim+1):(t-1))` needs the range to be splatted: `((previous_transform_dim+1):(t-1))...`. (Note: the "Batch 3D (in 5D)" test in `complex.jl` only tests regions where all batch dims fall into a single contiguous gap, so this path is never hit by existing tests.)
> 
> ### Bug 2: `circshift!` restore corrupts input and produces wrong results for rfft with external batch stride > 1
> 
> The shift amount is `-to_skip_x` but the restore is hardcoded to `+1` (`fft.jl:394` vs `fft.jl:413`). When the external batch dimension is not dim 1 (stride > 1), the restore doesn't undo the shift:
> 
> ```julia
> julia> x = CUDA.rand(Float32, 5, 3, 7, 4); xh = Array(x);
> 
> julia> y = rfft(x, (1, 3));
> 
> julia> maximum(abs.(Array(y) .- rfft(xh, (1, 3))))
> 21.903082   # should be ~1e-6
> 
> julia> maximum(abs.(Array(x) .- xh))
> 0.9303278   # input array permanently corrupted
> ```
> 
> Complex-to-complex transforms are unaffected (the circshift path only triggers for `eltype <: Real`).
> 
> ### Bug 3: `unique()` in NTuple plan methods causes type parameter mismatch
> 
> The NTuple-accepting `plan_*` methods (`fft.jl:215,225,236,246`) call `unique(region)` which can shrink the tuple, but the `R` type parameter is already fixed:
> 
> ```julia
> julia> plan_fft!(CUDA.rand(ComplexF32, 4, 4, 4), (1, 2, 1))
> ERROR: MethodError
> ```
> 
> The generic dispatch already handles `unique` before calling the NTuple method, so these redundant `unique` calls in the NTuple methods should just be removed.
> 
> ### Minor: dead code
> 
> - `front_view` (`fft.jl:175-178`): defined but never called after `unsafe_execute_trailing!` was removed.
> - `all_strides` (`wrappers.jl:74`): computed but never read.


Honestly though, the code in `cufftMakePlan` looks pretty messy... Here's what Claude had to say about that:

> The `inembed`/`cnembed` computation (lines 75-88 in `wrappers.jl`) is manually branched for nrank=1/2/3, but the branches follow a single pattern: `inembed[1]` is always `idist`, and the remaining entries are products of `input_size` between consecutive transform dims (reversed for C-order). This can be a one-liner:
> 
> ```julia
> inembed = Clonglong[idist, (prod(input_size[region[i]:region[i+1]-1]) for i in nrank-1:-1:1)...]
> cnembed = Clonglong[cdist, (prod(cdims[region[i]:region[i+1]-1]) for i in nrank-1:-1:1)...]
> ```
> 
> I verified this produces identical values for all cases (1D/2D/3D, contiguous and non-contiguous regions, various array sizes). Combined with removing dead code (`all_strides`, unused `res` binding) and cleaning up the stride computations, the entire `else` branch could be:
> 
> ```julia
>     else
>         internal_batch_dims, external_batch_dims = get_batch_dims(region, input_size)
> 
>         # Stride between consecutive elements in the innermost transform dim
>         istride = prod(input_size[1:region[1]-1])
>         # Distance between consecutive internal batches
>         idist = prod(input_size[1:internal_batch_dims[1]-1])
>         cdist = prod(cdims[1:internal_batch_dims[1]-1])
> 
>         # Embedded storage sizes (C-order): first entry >= n[0] (use idist),
>         # remaining entries are products of sizes between consecutive transform dims
>         inembed = Clonglong[idist, (prod(input_size[region[i]:region[i+1]-1]) for i in nrank-1:-1:1)...]
>         cnembed = Clonglong[cdist, (prod(cdims[region[i]:region[i+1]-1]) for i in nrank-1:-1:1)...]
> 
>         num_internal_batches = prod(input_size[collect(internal_batch_dims)])
> 
>         ostride = istride
>         # R2C: output uses half-complex layout; C2R: input uses half-complex layout
>         onembed, odist = input_type <: Real ? (cnembed, cdist) : (inembed, idist)
>         if output_type <: Real
>             inembed, idist = cnembed, cdist
>         end
> 
>         execution_type = promote_type(input_type, output_type)
>         cufftXtMakePlanMany(handle, nrank, Clonglong[rsz...],
>                             inembed, istride, idist, convert(cudaDataType, input_type),
>                             onembed, ostride, odist, convert(cudaDataType, output_type),
>                             num_internal_batches, worksize_ref, convert(cudaDataType, execution_type))
>     end
> ```

Sorry about the low-quality review, but it least it's better than nothing.

RainerHeintzmann:
Thanks a lot! I think this is of great help and (as always) shocking what bugs are apparently still in code that passes the tests. Did not know that Claude can do such good reviews. I will look at this in detail at some point.
Thanks a lot also for bringing this PR into sync with the main branch again. I was worried about how to do this.
Should I Sync the 3 non-synced commits before working on the fixes or better keep the current version?

RainerHeintzmann:
Quick question: How do I run the cufft   tests in the new arrangement? Is there a simple command?

maleadt:
> How do I run the cufft tests in the new arrangement? Is there a simple command?

It's a normal subpackage, so: `julia --project=lib/cufft -e 'using Pkg; Pkg.test()`
Best use Julia 1.12 for this.

> Should I Sync the 3 non-synced commits before working on the fixes or better keep the current version?

I'm not sure I understand; you have commits locally you didn't push? If so, better cherry-pick them on top of the merge commit I did here.

RainerHeintzmann:
from within the REPL I found that in the temp environment I can also do
```julia
include("lib/cufft/test/runtests.jl")
```
but for some strange reason I had to first add "AbtractFFTs.jl" into this environment.
 
As for my second question: It was not about my commits, but about new commits to the main branch of `CUDA.jl` that this PR was not synchronized to. But anyway, I will keep first fixing all the problems here, add test cases, simplify the code according to your suggestions, run tests and then sync with the main branch.

maleadt:
> from within the REPL I found that in the temp environment I can also do
> 
> ```julia
> include("lib/cufft/test/runtests.jl")
> ```
> 
> but for some strange reason I had to first add "AbtractFFTs.jl" into this environment.

You have to activate cuFFT's "test" environment, hence the `--project=lib/cufft/test`.



> As for my second question: It was not about my commits, but about new commits to the main branch of `CUDA.jl` that this PR was not synchronized to. But anyway, I will keep first fixing all the problems here, add test cases, simplify the code according to your suggestions, run tests and then sync with the main branch.

Just merge in `master`. Normally I'm more a fan of rebases, but at this point rebasing is impossible so you can keep merging.

RainerHeintzmann:
Thank a lot. This was really useful. I think all of those issues are fixed now. The branching was caused by legacy code, and you are right to simply it. However, I left a few more comments in the code, since this is a bit of a brain-twister and I think the comments may really help me or others to understand again how this all works.
I also added the bugs as test cases now. The question is a bit how to deal with dimensions such as (1,3,1), where there is still a slight difference in how fft() handles this for CuArray or standard Array. But there was this difference also before this PR.

lib/cufft/src/fft.jl
```diff
@@ -313,24 +341,92 @@ function unsafe_execute!(plan::CuFFTPlan{T,S,K,inplace}, x::DenseCuArray{S}, y::
...
+        end
+        # If there was at least one skip due to real Float32 alignment, we need to cyclicly rotate the whole array in place and run again
+        if (did_skip_x || did_skip_y)
+            if (did_skip_x)
+                circshift!((@view x[:]), -to_skip_x)
+            end
+            if (did_skip_y)
+                circshift!((@view y[:]), -to_skip_y)
+            end
+            n=1
```
maleadt: dead var?

lib/cufft/test/issues.jl
```diff
@@ -47,3 +47,22 @@ end
...
+    xc = copy(x)
+    y = rfft(x, (1, 3));
+    @test x == xc
+    @test maximum(abs.(Array(y) .- rfft(xh, (1, 3)))) < 1e-5
+end
+
+@testset "CUDA.jl_PR_issue_3" begin
+    # see https://github.com/JuliaGPU/CUDA.jl/pull/3052#issuecomment-4213439988
+    @test_throws ArgumentError plan_fft!(rand(ComplexF32, 4, 4, 4), (1, 2, 1)) 
+end
```
maleadt: Please give these descriptive testset names.

maleadt: Also, that `plan_fft!(rand(...))` is testing a CPU input

lib/cufft/src/fft.jl
```diff
@@ -130,23 +130,58 @@ function irfft(x::DenseCuArray{<:Union{Real,Integer,Rational}}, d::Integer, regi
...
+# it is convenient to use sort on tuples, but since this is only implemented as of Julia 1.12
+# and the dimensions are anyway limited to maximally three, we hand code the cases here:
+ensure_raising(sequence::NTuple{1, Int}) = sequence
+ensure_raising(sequence::NTuple{2, Int}) = (sequence[1] < sequence[2]) ? sequence : sequence[2:-1:1]
+ensure_raising(sequence::NTuple{3, Int}) = (sequence[1] < sequence[2]) ?
+            ((sequence[2]<sequence[3]) ? sequence : (sequence[1]<sequence[3]) ? sequence[[1,3,2]] : sequence[[3,1,2]]) :
+            ((sequence[1]<sequence[3]) ? sequence[[2,1,3]] : (sequence[2]<sequence[3]) ? sequence[[2,3,1]] : sequence[[3,2,1]])
+function ensure_raising(sequence::NTuple)
+    throw(ArgumentError("only up to three transform dimensions are allowed in one plan"))
 end
```
maleadt: `ensure_increasing`? You may also implement this using a tuple tail op (see Base) for something generic.

lib/cufft/src/fft.jl
RainerHeintzmann: Heck, yes. Thanks! ... removed.

lib/cufft/test/issues.jl
RainerHeintzmann: Thanks. Done. I also change `CUDA` to `CUDACore` where it seemed appropriate.

lib/cufft/src/fft.jl
RainerHeintzmann: Agreed. That is a better name and I changed it. Thanks! I am not sure how to use the `Base.tale(sequence)` here, though.

maleadt: `sequence` is only ever going to be 3 items, right? In that case, it's probably not worth using tuple peeling. But the current implementation allocates, so it may be better to do something like:
```julia
ensure_increasing(num::Number) = num
ensure_increasing(s::NTuple{1, Int}) = s
ensure_increasing(s::NTuple{2, Int}) = s[1] > s[2] ? (s[2], s[1]) : s
function ensure_increasing(s::NTuple{3, Int})
    s[1] > s[2] && (s = (s[2], s[1], s[3]))
    s[2] > s[3] && (s = (s[1], s[3], s[2]))
    s[1] > s[2] && (s = (s[2], s[1], s[3]))
    s
end
function ensure_increasing(::NTuple)
    throw(ArgumentError("only up to three transform dimensions are allowed in one plan"))
end
```

RainerHeintzmann: Great! Yes, I agree. I was not aware it was allocating and that this seemingly subtle difference made the difference. I updated using your code.

RainerHeintzmann:
Thanks again for the new edits and added tests!

PR #3057, opened 2026-03-16
Wrapper for  Blocksparse CuTensor code

kmp5VT:
Hi,

This is a wrapper type and functions to access the newly introduced blocksparse cutensor backend. Right now the code is expert level, i.e. users need to write a type that converts their object to CuTensorBS types or can achieve the low-level operations required by cutensor kernels.  I am still writing a test but the code is fully operational.

Thanks,
Karl


kmp5VT:
There were some issues in the Clang.jl's conversion of the cuTENSOR.h file into Julia wrapper functions. Specifically I had a runtime issue when trying to convert arrays of cuarray into `ptr{ptr{cvoid}}`. I think this is because CUDA.jl does not expect an array of cuarrays and so the julia side unsafe convert failed. This is not yet ready to merge.

kshyatt:
Thanks very much for putting this together, I'm happy to help with the header issues if needed!

kmp5VT:
@kshyatt I removed the extra code, made the functions that linked to the library relatively agnostic (i.e. you are not forced to use CuTensorBS but can buy in if you'd like) and added a unit test.  If you could help with the Clang.jl issue, that would be amazing!

kshyatt:
I'll try to take a look today!

kshyatt:
Did you use the scripts in `res/wrap` to do the wrapping of the C headers?

kmp5VT:
> Did you use the scripts in `res/wrap` to do the wrapping of the C headers?

Yes I did use the scripts but this produced the `Ptr{Ptr{Cvoid}}` definition in the libcutensor.jl file returns the following error 
```julia
ERROR: MethodError: no method matching unsafe_convert(::Type{Ptr{Nothing}}, ::CuPtr{Nothing})
The function `unsafe_convert` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  unsafe_convert(::Type{Ptr{Nothing}}, ::LibGit2.GitBlame)
   @ LibGit2 ~/.julia/juliaup/julia-1.12.1+0.x64.linux.gnu/share/julia/stdlib/v1.12/LibGit2/src/types.jl:1096
  unsafe_convert(::Type{Ptr{Nothing}}, ::LibGit2.GitRevWalker)
   @ LibGit2 ~/.julia/juliaup/julia-1.12.1+0.x64.linux.gnu/share/julia/stdlib/v1.12/LibGit2/src/types.jl:1096
  unsafe_convert(::Type{Ptr{Nothing}}, ::LibGit2.GitDiffStats)
   @ LibGit2 ~/.julia/juliaup/julia-1.12.1+0.x64.linux.gnu/share/julia/stdlib/v1.12/LibGit2/src/types.jl:1096
  ...

Stacktrace:
  [1] Ref{Ptr{Nothing}}(a::Vector{CuPtr{Nothing}})
    @ Base ./refpointer.jl:166
  [2] cconvert
    @ ./refpointer.jl:178 [inlined]
  [3] macro expansion
    @ ~/.julia/dev/CUDA.jl/lib/cutensor/src/libcutensor.jl:545 [inlined]
  [4] (::cuTENSOR.var"#cutensorBlockSparseContract##0#cutensorBlockSparseContract##1"{…})()
    @ cuTENSOR ~/.julia/packages/GPUToolbox/JLBB1/src/ccalls.jl:34
  [5] retry_reclaim
    @ ~/.julia/packages/CUDA/Il00B/src/memory.jl:434 [inlined]
  [6] check
    @ ~/.julia/dev/CUDA.jl/lib/cutensor/src/libcutensor.jl:22 [inlined]
  [7] cutensorBlockSparseContract
    @ ~/.julia/packages/GPUToolbox/JLBB1/src/ccalls.jl:33 [inlined]
  [8] 
    @ cuTENSOR ~/.julia/dev/CUDA.jl/lib/cutensor/src/blocksparse/operations.jl:50
  [9] contract!(alpha::Number, A::Any, Ainds::Vector{…}, opA::cuTENSOR.cutensorOperator_t, B::Any, Binds::Vector{…}, opB::cuTENSOR.cutensorOperator_t, beta::Number, C::Any, Cinds::Vector{…}, opC::cuTENSOR.cutensorOperator_t, opOut::cuTENSOR.cutensorOperator_t; jit::cuTENSOR.cutensorJitMode_t, workspace::cuTENSOR.cutensorWorksizePreference_t, algo::cuTENSOR.cutensorAlgo_t, compute_type::Nothing, plan::Nothing)
    @ cuTENSOR ~/.julia/dev/CUDA.jl/lib/cutensor/src/blocksparse/operations.jl:25
 [10] mul!(C::CuTensorBS{Float64, 1}, A::CuTensorBS{Float64, 3}, B::CuTensorBS{Float64, 2}, α::Float64, β::Float64)
    @ cuTENSOR ~/.julia/dev/CUDA.jl/lib/cutensor/src/blocksparse/interfaces.jl:21
```
However, I found that If I modify the code to be `Ptr{CuPtr{CVoid}}` that the blocksparse functionality works as expected with no error in either julia or C. This makes the function look closer to the `cutensorContract` function. Do you why clang.jl doesn't properly write these as `Ptr{CuPtr{CVoid}}`?

kshyatt:
Probably you missed some of the weird esoterica in `res/wrap`, haha. I'll fix it and make a PR to your PR?

kmp5VT:
@kshyatt I see your changes and this is useful for future reference. Thank you for the assistance! @lkdvos does this code work for you? Otherwise it is good to push on my end. @kshyatt let me know if there is anything I need to modify in the code

kshyatt:
Thanks for doing all the work to get this going, I think it will be quite useful for a bunch of TN packages...

lib/cutensor/src/blocksparse/interfaces.jl
```diff
@@ -0,0 +1,28 @@
...
+#     tC = promote_type(eltype(A), eltype(B))
+#     A_uniqs = [(idx, i) for (idx, i) in enumerate(A.inds) if !(i in B.inds)]
+#     B_uniqs = [(idx, i) for (idx, i) in enumerate(B.inds) if !(i in A.inds)]
+#     A_sizes = map(x->size(A,x[1]), A_uniqs)
+#     B_sizes = map(x->size(B,x[1]), B_uniqs)
+#     A_inds = map(x->x[2], A_uniqs)
+#     B_inds = map(x->x[2], B_uniqs)
+#     C = CuTensor(CUDA.zeros(tC, Dims(vcat(A_sizes, B_sizes))), vcat(A_inds, B_inds))
+#     return mul!(C, A, B)
+# end
```
lkdvos:
```suggestion
```

This is probably also left-over code

lib/cutensor/src/blocksparse/types.jl
```diff
@@ -0,0 +1,120 @@
...
+Base.show(io::IO, desc::CuTensorBSDescriptor) = @printf(io, "CuTensorBSDescriptor(%p)", desc.handle)
+
+Base.unsafe_convert(::Type{cutensorBlockSparseTensorDescriptor_t}, obj::CuTensorBSDescriptor) = obj.handle
+
+function unsafe_destroy!(obj::CuTensorBSDescriptor)
+    cutensorDestroyBlockSparseTensorDescriptor(obj)
+end
+
+## Descriptor function for CuTensorBS type. Please overwrite for custom objects
+function CuTensorBSDescriptor(A::CuTensorBS)
```
lkdvos, line 114: Is the distinction between `Int64`, `Int` and `Int32` here intentional?

lib/cutensor/src/blocksparse/types.jl
```diff
@@ -0,0 +1,120 @@
...
+        block_list = vcat(block_list, block...)
+    end
+    return block_list
+end
+
+# ## descriptor
+mutable struct CuTensorBSDescriptor
+    handle::cutensorBlockSparseTensorDescriptor_t
+    # inner constructor handles creation and finalizer of the descriptor
+    function CuTensorBSDescriptor(
```
lkdvos, line 72: I think it would be both helpful for clarity/self-documentation and for avoiding hard to decypher errors to restrict the types of these arguments in the inner constructor. This would also be more in line with the `CuTensorDescriptor` type + constructors.

lib/cutensor/src/blocksparse/types.jl
```diff
@@ -0,0 +1,120 @@
...
+end
+
+function CuTensorBSDescriptor(
+    numModes,
+    numNonZeroBlocks,
+    numSectionsPerMode,
+    extent,
+    nonZeroCoordinates,
+    eltype)
+
```
lkdvos, line 101: Perhaps a comment here to indicate which argument is filled in as `C_NULL` and what that means might be helpful

lib/cutensor/src/blocksparse/types.jl
lkdvos: [EDIT] just checked the cuTENSOR docs and indeed this seems correct.

lib/cutensor/src/blocksparse/types.jl
```diff
@@ -0,0 +1,120 @@
...
+function unsafe_destroy!(obj::CuTensorBSDescriptor)
+    cutensorDestroyBlockSparseTensorDescriptor(obj)
+end
+
+## Descriptor function for CuTensorBS type. Please overwrite for custom objects
+function CuTensorBSDescriptor(A::CuTensorBS)
+    numModes = Int32(ndims(A))
+    numNonZeroBlocks = Int64(length(A.nonzero_block_coords))
+    numSectionsPerMode = collect(Int32, A.blocks_per_mode)
```
lkdvos: If this has to be `Int32`, would it not be easier to immediately make that type restriction in the `CuTensorBS` type?

lib/cutensor/src/blocksparse/types.jl
```diff
@@ -0,0 +1,120 @@
...
+    extent = block_extents(A)
+    nonZeroCoordinates =  Int32.(vcat([[x...] for x in A.nonzero_block_coords]...) .- 1)
```
lkdvos: Same comment here. It also seems slightly strange to me to have a different storage format from the type that is required for the contraction, as this seems to introduce some allocations that could possibly be avoided?

lib/cutensor/src/blocksparse/types.jl
```diff
@@ -0,0 +1,120 @@
...
+    st = strides(A)
+    dataType = eltype(A)#convert(cuTENSOR.cutensorDataType_t, eltype(A))
+
+    ## Right now assume stride is NULL. I am not sure if stride works, need to discuss with cuTENSOR team.
```
lkdvos, line 125: Can we add an assert that the strides are the "natural ones" for that in the meantime?

lkdvos (commented):
Left some remaining comments, but for me I think most of the parts that I would use are there, since I don't really see myself going through the `CuTensorBS` construction (we also never used the `CuTensor` in TensorOperations so that is completely fine)

lib/cutensor/src/blocksparse/types.jl
kmp5VT: Marking this as resolved for now and fixing the other comments now. Thanks!

lib/cutensor/src/blocksparse/types.jl
kmp5VT: I added a dummy input comment to show that strides=C_NULL and a comment.

lib/cutensor/src/blocksparse/types.jl
kmp5VT: This makes sense. I added datatypes here for reference. I did have trouble mapping `C_NULL` to a datatype for the union so I just added a comment here and accept `Any` for now

lib/cutensor/src/blocksparse/types.jl
kmp5VT: Yes I move the restriction to CuTensorBS into the constructor

lib/cutensor/src/blocksparse/types.jl
kmp5VT: I think this is the natural stride i.e. all blocks are packed into a single list or address of the pointer to the start of block n is length(block_{n-1}) away from the address of the pointer to the start of block n-1

kmp5VT:
@kshyatt ~~In the tests, I skip the blocksparse test on CUDA versions that have a bug in the C library.~~
The C side error happens on the CUDA version that I know works. I am wondering if it is related to the cutensor version being picked up. Looking into this now.

lib/cutensor/src/blocksparse/operations.jl
```diff
@@ -0,0 +1,104 @@
...
+    # cuTENSOR expects a host-accessible array of GPU pointers
+    A_ptrs = CuPtr{Cvoid}[pointer(block) for block in A]
+    B_ptrs = CuPtr{Cvoid}[pointer(block) for block in B]
+    C_ptrs = CuPtr{Cvoid}[pointer(block) for block in C]
+    
+    cutensorBlockSparseContract(handle(), plan, 
+                                            Ref{scalar_type}(alpha), A_ptrs, B_ptrs, 
+                                            Ref{scalar_type}(beta),  C_ptrs, C_ptrs, 
+                                            plan.workspace, sizeof(plan.workspace), stream())
+    synchronize(stream())
```
kshyatt: Do we want to force synchronize here, or move this to the tests?

lib/cutensor/src/blocksparse/operations.jl
```diff
@@ -0,0 +1,104 @@
...
+    descA, modeA, opA,
+    descB, modeB, opB,
+    descC, modeC, opC,
+    descC, modeC, actual_compute_type)
+
+    plan_pref = Ref{cutensorPlanPreference_t}()
+    cutensorCreatePlanPreference(handle(), plan_pref, algo, jit)
+
+    plan = CuTensorPlan(desc[], plan_pref[]; workspacePref=workspace)
+    # cutensorDestroyOperationDescriptor(desc[])
```
kshyatt: Should this be uncommented?

lib/cutensor/test/contractions.jl
```diff
@@ -188,4 +188,69 @@ end
...
+eltypes_compact = [
+    (Float32, Float32, Float32, Float32),
+    (ComplexF32, ComplexF32, ComplexF32, Float32),
+     (Float64, Float64, Float64, Float64),
+     (ComplexF64, ComplexF64, ComplexF64, Float64)
+]
+## There are runtime issues in the cuTENSOR C library for versions 13.1, 12.6 and 12.0. 
+## These errors are fixed in other versions.
+if !(CUDACore.runtime_version() ∈ [
+    v"13.2", v"13.1",
```
kshyatt: This list seems inconsistent with the comment above

lib/cutensor/src/blocksparse/operations.jl
kmp5VT: There was a bug in the cuTENSOR C++ code that was causing an internal segfault to throw if the descriptor was destroyed too early. I can check to see if this is still there

lib/cutensor/test/contractions.jl
kmp5VT: I am fighting with CUDA + my cluster to figure out where cuTENSOR actually doesn't work this comment is out of date. Errors seem inconsistent on CI. For example I am testing with cuda version 13.0 but cutensor failed a 13.0 test on CI recently.

lib/cutensor/src/blocksparse/operations.jl
kmp5VT: This was probably an artifact of previous debugging. I believe it can be removed. Will check

kmp5VT: I checked this can be removed. Thanks!

lib/cutensor/src/blocksparse/operations.jl
kmp5VT: I just checked on cuda 12.6, this can be safely destroyed so I will uncomment

lib/cutensor/test/contractions.jl
kmp5VT: I think this problem was related to grabbing the wrong cutensor version. CI now passes for all versions so I removed this code.

kshyatt (approved):

PR #3058, opened 2026-03-22
Split CUDA.jl into several packages

maleadt:
I was finally fed up with CUDA.jl loading so slowly, most of which is caused by a large amount of methods being inserted (i.e., hard to fix), so I gave a shot trying to split off modules like CUBLAS into their own package. I was originally wanting to wait until Base figured out a way to split out LinearAlgebra/Random, since we're dealing with much of the same problems here, but since that is taking a long time I wanted to give it a go already.

Users will now have to actively import packages like cuBLAS or cuRAND to get the functionality they need:
- cuBLAS: matrix algebra
- cuSOLVER: factorizations
- cuFFT: FFTs
- cuRAND: random number generation
- cuSPARSE: sparse arrays

Some of these would have been possible with package extensions, others wouldn't, so I decided to fully go split packages.

The usual problematic areas are:
- `CUDA.rand`: remains, but only implemented after importing cuRAND.jl (`MethodError` otherwise, not sure how to improve)
- matrix multiplication etc: scalar iteration when not importing cuBLAS.jl

The idea would also be to version all these packages in lockstep, and maybe even extend this to cuDNN/cuTENSOR etc to create some kind of "CUDA.jl distribution". There were talks of adding this to Julia at some point, but we can do it manually for now.


vchuravy:
> Since this relies on [sources] for most testing, I'm considering bumping Julia compat to 1.11.

Oof Julia 1.10 is still the fastest Julia version for HPC use-cases (before Memory) and so this will create some tension, where I want to keep Julia 1.10 support, but also not lose out on modern CUDA.jl

kshyatt:
It's funny that we're recreating the `CuArrays.jl` + friends system we had many years ago, truly the old thing is new again (but I don't think this is a bad idea!).

kshyatt:
To @vchuravy's point, will we have a way to backport fixes/features onto a 1.10 compatible branch?

maleadt:
Hmm, looks like the improvements here are rather limited, at least after adding a precompilation workload. Before:

```
Benchmark 1: julia --project -e "using CUDA"
  Time (mean ± σ):      1.887 s ±  0.009 s    [User: 2.424 s, System: 0.383 s]
  Range (min … max):    1.872 s …  1.900 s    10 runs

Benchmark 2: julia --project -e "using CUDA; @cuda identity(nothing)"
  Time (mean ± σ):      2.531 s ±  0.031 s    [User: 3.105 s, System: 0.347 s]
  Range (min … max):    2.486 s …  2.580 s    10 runs
```


After:

```
Benchmark 1: julia --project -e "using CUDA"
  Time (mean ± σ):      1.415 s ±  0.008 s    [User: 1.960 s, System: 0.375 s]
  Range (min … max):    1.401 s …  1.428 s    10 runs

Benchmark 2: julia --project -e "using CUDA; @cuda identity(nothing)"
  Time (mean ± σ):      2.173 s ±  0.013 s    [User: 2.777 s, System: 0.313 s]
  Range (min … max):    2.153 s …  2.201 s    10 runs
```

Looking at `@time_imports`, it does reduce the CUDA.jl import time from 700ms to 200ms, so an impressive 60% reduction, but I'm not sure if it's worth the breakage.

kshyatt:
Maybe this is something we should revisit then when a new LTS is crowned, which will presumably support `[sources]`?

maleadt:
I wouldn't defer merging because of that; LTS isn't that important IMO, and with CompilerCaching being 1.11+ only I envision other parts of the stack dropping 1.10 too soon-ish. The larger breakage comes from the split itself, and this being a PR that will go stale instantly (while still taking a fair bit of time to get in reasonable shape) I'd rather we decide quickly whether to go with this or not.

maleadt:
That said, the 1.11 requirement is only for testing. Once things are registered, everything ought to work.

kshyatt:
Originally there was indeed a separate CUBLAS.jl package (some of my first commits in Julia...), will we need to do a registry transfer or something?

maleadt:
I don't find those packages back in General.

giordano:
> That said, the 1.11 requirement is only for testing. 

If that's the only motivation, then adding steps like https://github.com/EnzymeAD/Enzyme.jl/blob/f869ab339384754da7b344879baae2cadeae5c7a/.github/workflows/CI.yml#L117-L125 to the CI scripts should solve the problem, right? It's slightly annoying, and needed only for 1.10, but doesn't seem to be a deal breaker

maleadt:
So one reason I'm personally not 100% convinced myself about this PR is that we _may_ be able to instead tie all the fancy expensive functionality to Julia stdlibs using package extensions, i.e., load all of CUBLAS when importing LinearAlgebra.jl. However:
- that doesn't seem like it's happening soon?
- it would make the package extensions absolutely massive
- CUDA.jl would keep being massive in terms of artifacts (while with the proposed split here we could try and split up CUDA_Runtime_jll at some point)

maleadt:
Restored 1.10 testing. Do note that subpackage testing similarly requires 1.12 (workspaces), but otherwise doesn't impact the package. And we test the core of CUDA.jl first on all Julia versions.

kshyatt:
Do we export anything from CUBLAS? You can't export a symbol from a package extension

maleadt:
> You can't export a symbol from a package extension

Right, I guess that's another argument in favor of a strong split.

kshyatt:
It would be a huge problem for any extension with custom types e.g. `CUSPARSE`

maleadt:
I'll merge this when green.

The design has been changed, CUDA.jl is now a shim re-exporting everything from CUDACore so that this change in theory shouldn't be breaking. I did bump all versions to 6.0 though so that all versions are aligned now.

PR #3066, opened 2026-03-30
Mark `@device_override` as public

giordano:


maleadt (commented):
Can you bundle these at the top as a `@public foo, bar, baz` with `[ci skip]`?

giordano:
Done

PR #3074, opened 2026-04-01
Fix compat in profile

vchuravy:


maleadt (approved):
Doesn't it need an additional empty line too?

PR #3078, opened 2026-04-02
Add override for muladd and use LLVM intrinsic for fma

vchuravy:
Since Julia 0.7 (https://github.com/JuliaLang/julia/pull/22262) we are emitting

`muladd(a,b,c)` not as `llvm.fmuladd`, but rather as a sequence of:
```
%t = fmul contract %a %b
%r = fadd contract %t %c
```

The reason for that is vectorization of a potential reduction
(which is something we ought to investigate in Base if it is still worthwhile).

@efaulhaber had an example where

```
h = ...
for ...
  distance ...
  muladd(epsilon, h^2, distance^2)
```

and LLVM helpfully performed some code motion:

```
h = ...
t = mul(epsilon, h^2)
for ...
   add(t, distance^2)
```

Leading to a torn `contract` pair

```llvm
%69 = fmul contract float %"f::#parallel_foreach##10#parallel_foreach##11.fca.0.4.5.2.extract", %68
br label %L619, !dbg !197
;...
 %141 = fadd contract float %69, %140, !dbg !459
```

Manually using a fma leads to a performance improvement of `2.908` to `2.765` so ~5% faster.

I believe that the motivation in Base is not valid on GPUs since we benefit much more from the emission of fma,
in contrast to reduction vectorization.

@maleadt do you recall why we are using `__nv_fma`? LLVM should be able to perform better optimization over the `llvm.fma` intrinsic.

### DIFF
--- a/test/core/codegen.jl
+++ b/test/core/codegen.jl
@@ -22,6 +22,31 @@ end
     @test !occursin("@__nv_fmaf", ir)
 end
 
+@testset "fma uses LLVM intrinsic" begin
+    function fma_kernel(ptr)
+        unsafe_store!(ptr, fma(unsafe_load(ptr), unsafe_load(ptr,2), unsafe_load(ptr,3)))
+        return
+    end
+
+    for (T, suffix) in ((Float32, "f32"), (Float64, "f64"), (Float16, "f16"))
+        ir = sprint(io->CUDA.code_llvm(io, fma_kernel, Tuple{Ptr{T}}))
+        @test occursin("llvm.fma.$suffix", ir)
+        @test !occursin("__nv_fma", ir)
+    end
+end
+
+@testset "muladd uses LLVM intrinsic" begin
+    function muladd_kernel(ptr)
+        unsafe_store!(ptr, muladd(unsafe_load(ptr), unsafe_load(ptr,2), unsafe_load(ptr,3)))
+        return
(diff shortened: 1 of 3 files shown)


maleadt:
> @maleadt do you recall why we are using `__nv_fma`? LLVM should be able to perform better optimization over the `llvm.fma` intrinsic.

Sorry, no. This was probably an artifact of parsing what's in libdevice and wrapping everything, instead of a specific choice.
That said, doesn't `__nv_fma` expand to LLVM's FMA after libdevice inlining?

test/core/codegen.jl
```diff
@@ -22,6 +22,31 @@ end
...
+        unsafe_store!(ptr, muladd(unsafe_load(ptr), unsafe_load(ptr,2), unsafe_load(ptr,3)))
+        return
+    end
+
+    for (T, suffix) in ((Float32, "f32"), (Float64, "f64"), (Float16, "f16"))
+        ir = sprint(io->CUDA.code_llvm(io, muladd_kernel, Tuple{Ptr{T}}))
+        @test occursin("llvm.fmuladd.$suffix", ir)
+    end
+end
+
```
maleadt, line 49: Maybe test that the generated PTX code contains the expected instructions?

maleadt (approved):

PR #3080, opened 2026-04-02
Add preference to use CUDA_Runtime_Discovery for the compiler binaries

apozharski:
This works for my use case of ahead of time compilation, but now that I think about it may break if you use a local runtime and the artifact compiler, which I need to check, as well as not having any checks for compatibility of the compiler version and runtime version. It also might need some docs. As such, keeping this as [wip] for now.


apozharski:
The above concerns about mixing the local compiler and local runtime preferences has been addressed and everything should work correctly now.

CUDACore/src/CUDACore.jl
```diff
@@ -49,6 +47,15 @@ else
 end
 
 import Preferences
+const local_compiler = Preferences.@load_preference("local_compiler", "false") == "true"
+
+if local_compiler
+    using CUDA_Runtime_Discovery
+    const CUDA_Compiler = CUDA_Runtime_Discovery
+else
+    using CUDA_Compiler_jll
```
maleadt: Could you make it an `import` so that we're sure everything has been `CUDA_Compiler`-prefixed>

maleadt (dismissed):
Needs a bump in Project.toml for CUDA_Runtime_Discovery

maleadt:
Compat needs to be bumped in subpackages as well.

apozharski:
Hmmm it seems that a `GPUArrays` test fails consistently in CI and I can't repro it locally :thinking:

maleadt:
Could you report use of a local compiler in the `versioninfo()` output like we do for local toolkits?

And the failure is probably because you need to rebase on latest master in order to include https://github.com/JuliaGPU/CUDA.jl/pull/3088.

apozharski:
Added additional info for compiler in `versioninfo` (exactly same way with additional line as for the runtime, let me know if you want something different).

CUDATools/src/utilities.jl
```diff
@@ -22,6 +22,11 @@ function versioninfo(io::IO=stdout)
     end
     println(io, " for $(driver_version().major).$(driver_version().minor)")
     println(io, "- compiler $(compiler_version().major).$(compiler_version().minor)")
+    if CUDACore.local_compiler
```
maleadt, line 25: Preceding line needs to end with `, ` then

maleadt (approved):

Issue #3061, opened 2026-03-25
[CUSPARSE] Missing appropriate `similar` methods


alonsoC1s:
**Describe the bug**

Creating CUDA arrays with `similar` with a sparse matrix type, such as `SparseMatrixCSC`, works as expected when the number of dimensions is <=2. When there are 3 dimensions or more, the fallback for `Base.Array` is called. I understand that the CSC format only makes sense for 2d arrays. The proposed solution is to fall back to a `CuArray` constructor so the resulting array created with `similar(A <: CUSPARSE.AbstractCuSparseArray)` is still a GPU array. This prevents inadvertent scalar indexing down the line

**Expected behavior**
Array created with `similar` should be a GPU array
```julia
julia> using CUDA, SparseArrays
julia> A = sprand(20, 20, 0.5) |> cu
julia> A_sim_3d = similar(A, 20, 20, 5)
julia> typeof(A_sim_3d)
Array{Float32, 3} # Should be of type `CuArray` or similar
```

```
CUDA toolchain: 
- runtime 13.1, artifact installation
- driver 535.261.3 for 13.2
- compiler 13.2

CUDA libraries: 
- CUBLAS: 13.2.1
- CURAND: 10.4.1
- CUFFT: 12.1.0
- CUSOLVER: 12.0.9
- CUSPARSE: 12.7.3
- CUPTI: 2025.4.1 (API 13.1.1)
- NVML: 12.0.0+535.261.3

Julia packages: 
- CUDA: 5.10.1
- GPUArrays: 11.4.1
- GPUCompiler: 1.8.2
- KernelAbstractions: 0.9.41
- CUDA_Driver_jll: 13.2.0+0
- CUDA_Compiler_jll: 0.4.2+0
- CUDA_Runtime_jll: 0.20.1+0

Toolchain:
- Julia: 1.12.1
- LLVM: 18.1.7

1 device:
  0: NVIDIA GeForce RTX 2060 (sm_75, 4.841 GiB / 6.000 GiB available)
```

If there is no PR for this, I would be happy to attempt one

maleadt:
> The proposed solution is to fall back to a `CuArray` constructor so the resulting array created with `similar(A <: CUSPARSE.AbstractCuSparseArray)` is still a GPU array.

That seems like a good idea, especially as it aligns with Array:

```julia-repl
julia> using SparseArrays

julia> A = sprand(20, 20, 0.5)
[...]

PR #3091, opened 2026-04-10
cuSPARSE: Align default behavior of higher-dimensional similar with SparseArrays

alonsoC1s:
As mentioned in #3061 , when calling `similar` with CUSPARSE arguments and dimensions higher than 2, the fallback CPU method is called. As a result, calls to `similar` with GPU (sparse) matrices sometimes return host arrays, causing scalar indexing issues down the line

This PR introduces a fallback to `similar` that creates un-initialized `CuArray`s when there are 3 or more dimensions given as arguments, which is similar to what Base and SparseArrays do

Important note: I wasn't able to fully test this locally due to some issues instantiating the repo environment with Pkg

### DIFF
--- a/lib/cusparse/src/array.jl
+++ b/lib/cusparse/src/array.jl
@@ -420,16 +420,86 @@ Base.getindex(A::AbstractCuSparseMatrix, i, ::Colon)       = getindex(A, i, 1:si
 Base.getindex(A::AbstractCuSparseMatrix, ::Colon, i)       = getindex(A, 1:size(A, 1), i)
 Base.getindex(A::AbstractCuSparseMatrix, I::Tuple{Integer,Integer}) = getindex(A, I[1], I[2])
 
-# Indexing paths that produce a 1-D sparse result are implemented via a host-side
-# round-trip through `SparseMatrixCSC`. This keeps `Base._unsafe_getindex` out of
-# the picture: it would otherwise call `setindex!` on the `CuSparseVector` returned
-# by `similar`, which we don't define.
-# TODO: implement directly on the device. For CSC the linearization is mechanical
-# (no sort needed); CSR/COO/BSR need either a sort or a CSC conversion first.
-Base.getindex(A::CuSparseMatrix, ::Colon)                            = CuSparseVector(SparseMatrixCSC(A)[:])
-Base.getindex(A::CuSparseMatrix, I::AbstractUnitRange)               = CuSparseVector(SparseMatrixCSC(A)[I])
-Base.getindex(A::CuSparseMatrix, i::AbstractUnitRange, j::Integer)   = CuSparseVector(SparseMatrixCSC(A)[i, j])
-Base.getindex(A::CuSparseMatrix, i::Integer, j::AbstractUnitRange)   = CuSparseVector(SparseMatrixCSC(A)[i, j])
+# Indexing paths that produce a 1-D sparse result. These have to bypass
+# `Base._unsafe_getindex` — it would otherwise call `setindex!` on the
+# `CuSparseVector` returned by `similar`, which we don't define.
+
+# Linearization: write `(j-1)*nrows + rowVal[k]` per entry. CSC storage order is
+# already column-major with sorted rows within each column, so the output `iPtr` is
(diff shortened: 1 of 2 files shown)


maleadt:
Test failures related.

alonsoC1s:
@maleadt my bad. I think I got it this time

alonsoC1s:
I found what was making the tests fail. I think there are a few workarounds, so I'd like some opinions on which is more suitable

Basically, I defined the signature `Base.similar(::CuSparseMatrix, T::Type, dims::NTuple{N, Int}) where N` thinking the cases of tuples of length 1 and 2 were covered, and this only dispatched for 3-tuples or longer. However, this ended up being called instead of the more general `Base.similar(A::AbstractArray, I::Tuple{Int})`, which is used inside `Base._unsafe_getindex`. Effectively, this broke the implicit linearization that happens when you do:
```julia
A[:]
```

I tried easy fixes with limited success. A few possibilities to do this correctly would be
1. Explicitly figure out the linearization so that it works as Base is expecting
2. Only patch the fallback to `Base.similar` when 3 dimensions (as opposed to the normal 2) are given
3. Patch the fallback only for `N >= 3` by using generated functions or something similar

I'm not sure what the best way to proceed would be @maleadt

maleadt:
Can we copy SparseArrays.jl and only provide `similar` in CUSPARSE for only 1/2D? https://github.com/JuliaSparse/SparseArrays.jl/blob/3ab784c88070f1e8d4c3e424c9297547a956f63f/src/sparsematrix.jl#L723-L724
And then provide a `similar(a::AbstractGPUSparseArray, ::Type{T}, dims::Dims{N})` fallback that allocates a `CuArray`?

alonsoC1s:
If I'm understanding correctly, this is what I tried to implement at first and what ended up causing problems down the line in Base. The fallback with `Dims{N}` (I used `Ntuple` directly, but this is better) ended up inadvertently covering the case when similar is called for a CUSPARSE array with a 1-tuple as `dims`. This particular method shouldn't be overloaded, as Base relies on it when doing implicit linearization

The problem isn't catching the case SparseArrays catches with `Dims{N}`, the problem is that doing so would take precedence over another method, and break lots of things. Defining 
```julia
similar(::AbstractGPUSparseArray, ::Type{T}, dims::Dims{1})
```
ourselves would need to somehow call the much more general method for AbstractArrays with the right arguments. The ideal case would be to dispatch on `Dims{N}` if and only if `N >= 3` while leaving `N=1` alone

lib/cusparse/src/array.jl
```diff
@@ -288,18 +288,25 @@ Base.similar(Mat::CuSparseMatrixCOO, T::Type) = CuSparseMatrixCOO(copy(Mat.rowIn
...
 Base.similar(Mat::CuSparseMatrixCSR{Tv, Ti}, N::Int, M::Int) where {Tv, Ti} = similar(Mat, Tv, N, M)
 Base.similar(Mat::CuSparseMatrixCOO{Tv, Ti}, N::Int, M::Int) where {Tv, Ti} = similar(Mat, Tv, N, M)
+# For dims higher than 3 we default to dense arrays as does Base
+Base.similar(::CuSparseMatrix{Tv}, dims::Vararg{Int, N}) where {N, Tv} = CuArray{Tv}(undef, dims)
 
 Base.similar(Mat::CuSparseMatrixCSC, T::Type, dims::Tuple{Int, Int}) = similar(Mat, T, dims...)
 Base.similar(Mat::CuSparseMatrixCSR, T::Type, dims::Tuple{Int, Int}) = similar(Mat, T, dims...)
 Base.similar(Mat::CuSparseMatrixCOO, T::Type, dims::Tuple{Int, Int}) = similar(Mat, T, dims...)
+# The next one would overwrite `similar` for dims::Tuple{Int}, which breaks linearization
+# Base.similar(::CuSparseMatrix, T::Type, dims::NTuple{N, Int}) where N = CuArray{T}(undef, dims)
```
alonsoC1s: This line works effectively the same as the fallback from SparseArrays. It uses `NTuple` instead of `Dims{N}`, but `Dims{N}` would have the same issue AFAICT

PR #3095, opened 2026-04-10
Add `_throw_dmrs` device override for reshape of views

Abdelrahman912:
## Problem

`reshape(@view(data[1:n*n]), (n, n))` fails to compile on the GPU. `@view` creates a `SubArray`, which has no specialized `reshape` method on the device, so it falls back to Base's generic `_reshape`. That path calls `_throw_dmrs`, which tries to construct a `DimensionMismatch` string — unsupported on the GPU.

```
Reason: unsupported call to an external C function
Stacktrace:
 [1] _string_n
   @ ./strings/string.jl:109
 [2] StringMemory
   @ ./iobuffer.jl:167
 [3] dec
   @ ./intfuncs.jl:918
 [4] #string#403
   @ ./intfuncs.jl:1000
 [5] multiple call sites
   @ unknown:0
   
  ```
  
  ```
  function _reshape(parent::AbstractArray, dims::Dims)
    n = length(parent)
    prod(dims) == n || _throw_dmrs(n, "size", dims)
    __reshape((parent, IndexStyle(parent)), dims)
end

@noinline function _throw_dmrs(n, str, dims)
    throw(DimensionMismatch("parent has $n elements, which is incompatible with $str $dims")) ## THIS IS THE CULPRIT
end
  ```

### DIFF
--- a/CUDACore/src/device/quirks.jl
+++ b/CUDACore/src/device/quirks.jl
@@ -103,3 +103,7 @@ for op in (:(<), :(<=), :cmp)
         @device_override Base.$op(q::Rational, x::AbstractFloat) = $op(float(q), x)
     end
 end
+
+# reshape.jl
+@device_override Base._throw_dmrs(n, str, dims) =
+    @gputhrow "DimensionMismatch" "Dimensions mismatch when reshaping. New dimensions must be consistent with array size"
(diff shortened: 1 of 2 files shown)


maleadt:
Please don't continuously merge `master`. That's a waste of CI resources.

maleadt (approved):

PR #3099, opened 2026-04-14
Improve performance of native RNG

### LINKED ISSUE #3056: Performance improvement ideas for randn! Float32
I created a package to help generate random numbers on the GPU https://github.com/medyan-dev/PhiloxRNG.jl

The method looks to be about 4x faster than CUDA.jl for filling a Float32 vector with randn! So there is room to improve the randn! kernel. I haven't looked deeply into why there is such a big difference in performance yet.

maleadt:
Rewrites the `NativeRNG` kernels to use stateless Philox4x32-10 with fast polynomial math, closing the performance gap with [PhiloxRNG.jl](https://github.com/medyan-dev/PhiloxRNG.jl).

Closes #3056.

## Motivation

PhiloxRNG.jl reported ~4x faster `randn!` Float32 compared to CUDA.jl. Investigation showed several issues in the `NativeRNG` kernels:

- **32 threads/block** — poor GPU occupancy
- **Shared-memory Philox2x32** — per-warp state management overhead, only 2 outputs per call
- **Standard `log`/`cos`/`sin`** in Box-Muller — expensive libm calls instead of polynomial approximations
- **`Float64(2)^(-64)`** — doesn't constant-fold on GPU, compiles to `__internal_accurate_pow` runtime calls

## Changes

- Replace device-side Philox2x32 with **stateless Philox4x32-10** (4 UInt32 outputs per call, no shared memory)
- Add **fast polynomial log** (fdlibm-based minimax) and **fast sincospi** (branchless octant-based minimax) for Box-Muller
- Use **256 threads/block**
- **4 Float32 / 2 Float64** values per Philox call
- Pre-compute Float64 power-of-two constants to avoid GPU `pow` calls
- Specialized kernels for Float32, Float64, Complex{Float32}, Complex{Float64}; generic fallback preserved for other types

### vs old NativeRNG

| Operation    | Before     | After       | Speedup |
|-------------|------------|-------------|---------|
| rand! F32   | 271 GB/s   | **920 GB/s**  | 3.4x    |
| randn! F32  | 536 GB/s   | **920 GB/s**  | 1.7x    |
| rand! F64   | 20.7 GB/s  | **916 GB/s**  | 44x     |
| randn! F64  | 12.4 GB/s  | **109 GB/s**  | 8.8x    |

### vs cuRAND vendor library

| Operation    | NativeRNG  | cuRAND lib  | Ratio   |
|-------------|------------|-------------|---------|
| rand! F32   | **920 GB/s** | 906 GB/s    | 1.02x   |
| randn! F32  | **920 GB/s** | 178 GB/s    | 5.2x    |
| rand! F64   | **916 GB/s** | 506 GB/s    | 1.8x    |
| randn! F64  | **109 GB/s** | 25.6 GB/s   | 4.3x    |

### vs PhiloxRNG.jl (the reference package from the issue)

| Operation    | NativeRNG  | PhiloxRNG.jl |
|-------------|------------|--------------|
| rand! F32   | 920 GB/s   | 920 GB/s     |
| randn! F32  | 920 GB/s   | 920 GB/s     |

Float32 operations are now fully memory-bandwidth bound on the RTX 5080 (~920 GB/s).

### DIFF
--- a/lib/curand/src/native.jl
+++ b/lib/curand/src/native.jl
@@ -89,8 +89,13 @@ end
     fma(Float32(u), Float32(2)^(-32), Float32(2)^(-33))
 end
 
+# Pre-computed constants to avoid Float64 ^ calls that don't constant-fold on the GPU
+const F64_2P_NEG64 = Float64(2)^(-64)
+const F64_2P_NEG65 = Float64(2)^(-65)
+const F64_2P_NEG66 = Float64(2)^(-66)
+
 @inline function u01(::Type{Float64}, u::UInt64)
-    fma(Float64(u), Float64(2)^(-64), Float64(2)^(-65))
+    fma(Float64(u), F64_2P_NEG64, F64_2P_NEG65)
 end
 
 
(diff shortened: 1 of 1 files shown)


maleadt:
Actually, with things being stateless we can consider moving it to GPUArrays...

PR #3101, opened 2026-04-15
CUSPARSE: fix COO scalar getindex crossing row boundaries (#3100)

### LINKED ISSUE #3100: [cuSPARSE] Incorrect indexing for COO-formatted sparse arrays
**Describe the bug**
Indexing the last column of a COO-formatted sparse matrix returns zeros as a result

**To reproduce**

```julia
using CUDA, SparseArrays
C = cuSPARSE.CuSparseMatrixCOO(sparse(reshape(1:16, 4, 4)))

julia> CUDA.@allowscalar C[1, 4]
0

julia> CUDA.@allowscalar C[2, 4]
0

julia> CUDA.@allowscalar C[4, 4]
16

julia> CUDA.@allowscalar C[1, 3]
9
```

<p>

```

```

</p>
</details>

**Expected behavior**

```julia
@assert CUDA.@allowscalar C[1, 4] == 13
```

**Version info**

Details on Julia:

```
Julia Version 1.12.6
Commit 15346901f00 (2026-04-09 19:20 UTC)
Build Info:
  Official https://julialang.org release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 8 × Intel(R) Core(TM) i7-9700K CPU @ 3.60GHz
  WORD_SIZE: 64
  LLVM: libLLVM-18.1.7 (ORCJIT, skylake)
  GC: Built with stock GC
Threads: 8 default, 1 interactive, 8 GC (on 8 virtual cores)
Environment:
  LD_LIBRARY_PATH = /software/CUDA/cuda-12.2/lib64:
  JULIA_NUM_THREADS = auto
```

Details on CUDA:

```
CUDA toolchain:
- runtime 13.2, artifact installation
- driver 535.261.3 for 13.2
- compiler 13.2

CUDA libraries:
- cuBLAS: 13.3.0
- cuSPARSE: 12.7.9
- cuSOLVER: 12.1.0
- cuFFT: 12.2.0
- cuRAND: 10.4.2
- CUPTI: 2026.1.0 (API 13.2.0)
- NVML: 12.0.0+535.261.3

Julia packages:
- CUDACore: 6.0.0
- GPUArrays: 11.4.1
- GPUCompiler: 1.9.1
- KernelAbstractions: 0.9.41
- CUDA_Driver_jll: 13.2.0+0
- CUDA_Compiler_jll: 0.4.2+0
- CUDA_Runtime_jll: 0.21.0+0

Toolchain:
- Julia: 1.12.6
- LLVM: 18.1.7

1 device:
  0: NVIDIA GeForce RTX 2060 (sm_75, 4.891 GiB / 6.000 GiB available)
```

**Additional context**

I believe the indexing bug comes from the assumptions of `searchsortedfirst` not being met when used to find the [column index](https://github.com/JuliaGPU/CUDA.jl/blob/a9a687cf541c1cdc91e0b6884382f6fd57ceb417/lib/cusparse/src/array.jl#L495). The fix probably requires changes to that single line and no more

maleadt:
Resolves #3100 plus a family of related correctness issues in scalar `getindex` across all cuSPARSE sparse types. The root cause was that every `getindex` method binary-searched within a major-axis slice, silently assuming two invariants that cuSPARSE doesn't actually guarantee: (1) minor indices are sorted within each major-axis slice, and (2) `(i, j)` entries are unique. Both invariants are routinely violated in practice.

Switched all five `getindex` methods (`CuSparseVector`, `CuSparseMatrix{CSC,CSR,COO,BSR}`) to linear-scan the major-axis slice and accumulate matching values, just like cuPy does.

### DIFF
--- a/lib/cusparse/test/array.jl
+++ b/lib/cusparse/test/array.jl
@@ -216,6 +216,12 @@
             @test d_coo[2, 2] == 0
             @test d_coo[1, 2] == 0
         end
+        let d_vec = CuSparseVector{Int}(CuArray(Int32[3, 1, 3]), CuArray([10, 7, -4]), 4)
+            @test d_vec[1] == 7
+            @test d_vec[3] == 6
+            @test d_vec[2] == 0
+            @test d_vec[4] == 0
+        end
         # Bool duplicates combine via OR, not integer sum.
         let d_bool = CuSparseMatrixCSR(
                 CuArray(Int32[1, 3, 3]),
--- a/lib/cusparse/test/array.jl
+++ b/lib/cusparse/test/array.jl
@@ -127,6 +127,25 @@
             @test Array(d_x[i, :]) == x[i, :]
         end
     end
+    # scalar getindex at every (i, j) — regression test for #3100,
+    # where COO column search crossed into the next row's entries
+    let dense = sparse(reshape(1:16, 4, 4)),
+        d_dense = CuSparseMatrixCOO(dense)
+        CUDACore.@allowscalar begin
+            for j in 1:4, i in 1:4
+                @test d_dense[i, j] == dense[i, j]
+            end
+        end
+    end
+    # sparse case with empty rows and missing entries
+    let s = sparse([1, 1, 3, 4], [1, 3, 2, 4], [10, 20, 30, 40], 4, 4),
+        d_s = CuSparseMatrixCOO(s)
+        CUDACore.@allowscalar begin
+            for j in 1:4, i in 1:4
+                @test d_s[i, j] == s[i, j]
+            end
+        end
+    end
     y = sprand(k,n,0.2)
     d_y = CuSparseMatrixCOO(y)
     @test_throws ArgumentError copyto!(d_y,d_x)
--- a/lib/cusparse/test/array.jl
+++ b/lib/cusparse/test/array.jl
@@ -159,6 +159,33 @@
             end
         end
     end
+    # CSR with unsorted column indices within each row — can arise from
+    # SpGEMM and other operations, so getindex must not binary-search.
+    let dense = sparse(reshape(1:16, 4, 4)),
+        d_scrambled = CuSparseMatrixCSR(
+            CuArray(Int32[1, 5, 9, 13, 17]),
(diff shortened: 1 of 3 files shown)


kshyatt:
Fun find through [git blame](https://github.com/JuliaGPU/CUDA.jl/commit/835373677e735b9d226b378a6a09b38f80acbac9)

lib/cusparse/test/array.jl
```diff
@@ -127,6 +127,25 @@
             @test Array(d_x[i, :]) == x[i, :]
         end
     end
+    # scalar getindex at every (i, j) — regression test for #3100,
+    # where COO column search crossed into the next row's entries
+    let dense = sparse(reshape(1:16, 4, 4)),
+        d_dense = CuSparseMatrixCOO(dense)
+        CUDACore.@allowscalar begin
+            for j in 1:4, i in 1:4
```
kshyatt: can we not hardcode these bounds? couldn't we just use `axes` for this inner loop?

maleadt:
Hmm, looking at the CI failure, I think we're going about it wrong here. AFAIU our `getindex` is assuming COO contents are sorted by row and column, while the docs only mention sorting by row. So we can't simply do a binary search, but have to do a linear scan.

alonsoC1s:
@maleadt I have tried patching it by changing
```julia
c1 = searchsortedfirst(A.colInd, i1, r1, r2, Base.Order.Forward)
```
to
```julia
c1 = searchsortedfirst(A.colInd, i1, r1, r2-1, Base.Order.Forward)
```

My understanding of the issue was that the binary search that's done inside `searchsortedfirst` is thrown off by the fact that the last element of `A.colInd[r1:r2]` isn't sorted. For instance, like `[1, 2, 3, 1]`. If we eliminate the last element, `searchsortedfirst` will do what it's supposed to do, no need to change the other checks outside of that specififc line. This is the way I implemented it on device arrays and from my testing, it works fine

maleadt:
> If we eliminate the last element, `searchsortedfirst` will do what it's supposed to do, no need to change the other checks outside of that specififc line. This is the way I implemented it on device arrays and from my testing, it works fine

COO is only documented as sorted by `row`, not by `(row, col)`, so I don't think this suffices.

PR #3103, opened 2026-04-15
Use Republic.jl to forward subpackage APIs

AntonOresten:
I recently created [Republic.jl](https://github.com/MurrellGroup/Republic.jl) to replace and extend [Reexport.jl](https://github.com/JuliaLang/Reexport.jl). After jumping the gun on v1 (which had questionable inheritance semantics) I'm now much more confident in v2, taking inspiration from the pre-1.11 `@public` machinery from #3058.

The previous code has special handling for [GPUToolbox.jl](https://github.com/JuliaGPU/GPUToolbox.jl)'s enums, so this PR relies on GPUToolbox switching to Republic.jl as well for pre-1.11 enum support.

This would be my first real stress-test of Republic, but I don't see any potential problems yet.


maleadt:
Interesting! I'd be fine with depending on Republic.jl for this. Any reason you went with a new package, and not an extension to Reexport.jl? Especially seeing the `s/@reexport/@republic reexport=true/`.

AntonOresten:
Reexport.jl has a very clear scope, and the `public` keyword introduces a different dimension — forwarding non-exported names, cross-version tracking, composing inheritance (implicit imports) with re-exporting. Bolting that onto Reexport felt like stretching it beyond its purpose. Also the name was too good to pass up.

There could definitely be `@reexport` -> `@republic reexport=true` for convenience though. I also think https://github.com/MurrellGroup/Republic.jl/issues/9 is a missing piece for expressiveness.

PR #3106, opened 2026-04-16
Export gputhrow

termi-official:
~~`@gputhrow` was exported on v5~~

### DIFF
--- a/CUDACore/src/device/quirks.jl
+++ b/CUDACore/src/device/quirks.jl
@@ -7,6 +7,8 @@
 # information about the exception type and reason in the `ExceptionInfo` struct, where it
 # will be read by the GPU runtime and printed to the console.
 
+@public @gputhrow
+
 macro gputhrow(subtype, reason)
     quote
         info = kernel_state().exception_info


maleadt:
It was not? https://github.com/JuliaGPU/CUDA.jl/blob/release-5.11/src/device/quirks.jl

termi-official:
Oh, right. Sorry. I was really just took `CUDA: @gputhrow` from somewhere and was assuming it was part of the public API.

We would like to do the following in our package: `@device_override @noinline Ferrite.throw_detJ_not_pos(detJ) = @gputhrow("ArgumentError", "det(J) is not positive. Please check the value on CPU.")` to still give reasonable error messages. Would it be possible to add `@gputhrow` to the public API or should we do something else?

maleadt:
The functionality is too much tied to internal state for it to become public, I think. Especially because I plan to rework exception handling, albeit at some undetermined point in the future. Would it be fine if you lose some of the reporting accuracy and simply do what we used to do before:
```
@cuprintln "ERROR: " $(args...) "."
throw(nothong)
```

termi-official:
I see. We will stick with whatever is currently recommended.

PR #3109, opened 2026-04-17
Reorganize and untangle subpackage tests (NFC)

maleadt:


kshyatt:
I feel like this is going to be really hard to review because it has 46 (!!!) changed files and a massive diff. Is there any way this could be split into multiple PRs, one per sub package?

maleadt:
Eh, I think this is always going to very hard to review. It's a bunch of code being moved around, affecting tests only, so should be NFC.

I mean, it's hard to split up something like https://github.com/JuliaGPU/CUDA.jl/pull/3109/changes/d80500b2d545d345e536841bf880133eaac66100, and GitHub is going to be awful to render if even if split up because of how it deals with file changes.

PR #3110, opened 2026-04-18
Switch to ParallelTestRunner

maleadt:
Needs https://github.com/JuliaTesting/ParallelTestRunner.jl/pull/129


maleadt:
51/41/37 mins here vs 42/34/33 on master for Julia 1.11/1.12/1.13. And this surprisingly looks like the actual tests slowing down, e.g. on Julia 1.12:

```
gpuarrays/linalg/core                          (9) │   152.89 │   0.02 │      28.29 │   148.00 │   4.48 │  2.9 │   13223.51 │  4035.78 │
gpuarrays/linalg/norm                         (13) │   281.17 │   0.02 │       0.03 │   130.00 │   7.46 │  2.7 │   17295.61 │  3769.99 │
```

vs

```
gpuarrays/linalg/core                         (3) |   126.39 |   0.02 |  0.0 |      28.29 |   226.00 |   3.43 |  2.7 |   11518.49 | 10666.16 |
gpuarrays/linalg/norm                         (4) |   250.23 |   0.02 |  0.0 |       0.03 |   146.00 |   6.00 |  2.4 |   14957.74 |  6904.70 |
```

giordano:
Try with `--verbose` which shows also the init time?

maleadt:
Good idea. It's probably related to my tuning of the memory pool heuristics though, and not because of PTR.

maleadt:
The only remaining issue is the multigpu one. I'm looking into this separately, but I think we can merge this already.

giordano:
What's the issue specifically?

maleadt:
> What's the issue specifically?

Test failure due to driver bug (presumably), unrelated to PTR but it does seem to make it occur more often.

PR #3113, opened 2026-04-21
Fix workspace size on 5.11 and CUDA 13.2, backport #3062

kshyatt:
https://github.com/JuliaGPU/CUDA.jl/pull/3062 fixed this for CUDA 6, but some packages can't update to that yet


kshyatt:
OK, looks like this worked!

PR #3119, opened 2026-04-29
Fix Int32 overflow in !range metadata for large intrinsics

### LINKED ISSUE #3024: Indexing range metadata is set up wrong

rainerrodrigues:
Fixes #3024

### Summary
This PR fixes a bug where `!range` metadata on LLVM intrinsics was causing an `Int32` overflow, which resulted in negative upper bounds and prevented LLVM from optimizing away standard bounds checks. 
### The Cause
LLVM `!range` metadata expects a half-open interval `[start, stop)`. In `CUDACore/src/device/intrinsics/indexing.jl`, the exclusive upper bound is calculated as `(range.stop + 1) % Int32`. 
For intrinsics with large maximums, like `gridDim_x` (`2^31 - 1`), adding 1 overflows the signed 32-bit integer, resulting in `-2147483648`. LLVM interprets this as a wrap-around range, effectively destroying bounds-check optimizations.
### The Fix
Added a safeguard in `_index` to only attach the `!range` metadata if `range.stop < typemax(Int32)`. Small-bounded intrinsics (like `blockDim`) keep their metadata, while large-bounded ones (like `gridDim`) safely omit it to prevent the overflow.
### Testing
Added a unit test in `test/core/device/intrinsics.jl` using `CUDA.code_llvm(..., dump_module=true, raw=true)` to verify that:
1. `!range` is correctly emitted for `blockDim().x`.
2. `!range` is safely omitted for `gridDim().x`.

### DIFF
--- a/lib/cusparse/test/construction.jl
+++ b/lib/cusparse/test/construction.jl
@@ -88,4 +88,28 @@ if capability(device()) >= v"5.3"
         end
     end
 end
+
+@testset "similar for COO and BSR with custom types/dims" begin
+    row = CuVector{Cint}([1, 2, 3])
+    col = CuVector{Cint}([1, 2, 3])
+    val = CuVector{Float32}([1.0, 2.0, 3.0])
+    coo_mat = CuSparseMatrixCOO(row, col, val, (3, 3))
+
+    coo_sim = similar(coo_mat, Float64, (3, 3))
+    @test eltype(coo_sim) == Float64
+    @test size(coo_sim) == (3, 3)
+    @test typeof(coo_sim) <: CuSparseMatrixCOO
+
+    rowPtr = CuVector{Cint}([1, 2, 3])
+    colVal = CuVector{Cint}([1, 2])
+    valBSR = CuVector{Float32}([1.0, 2.0, 3.0, 4.0]) 
+    bsr_mat = CuSparseMatrixBSR(rowPtr, colVal, valBSR, 2, 'R', 1, (4, 4))
+
+    bsr_sim = similar(bsr_mat, Float64, (4, 4))
+    @test eltype(bsr_sim) == Float64
+    @test size(bsr_sim) == (4, 4)
+    @test typeof(bsr_sim) <: CuSparseMatrixBSR
+    @test bsr_sim.blockDim == bsr_mat.blockDim
+end
+
 end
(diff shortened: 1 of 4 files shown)


maleadt:
CI failures look related.

maleadt:
What's with the unrelated cuSPARSE changes? Otherwise this PR looks fine.

rainerrodrigues:
Ah, sorry about that! My local Git branches got a bit tangled while I was trying to resolve a stubborn CI precompilation cache issue, and I accidentally pulled in a file from another sparse matrix PR I am working on.

I have just pushed a commit to revert the cuSPARSE changes, so this PR is now strictly focused on the LLVM range metadata fix. Thanks for catching that!

lib/cusparse/test/construction.jl
```diff
@@ -88,4 +88,28 @@ if capability(device()) >= v"5.3"
...
+    rowPtr = CuVector{Cint}([1, 2, 3])
+    colVal = CuVector{Cint}([1, 2])
+    valBSR = CuVector{Float32}([1.0, 2.0, 3.0, 4.0]) 
+    bsr_mat = CuSparseMatrixBSR(rowPtr, colVal, valBSR, 2, 'R', 1, (4, 4))
+
+    bsr_sim = similar(bsr_mat, Float64, (4, 4))
+    @test eltype(bsr_sim) == Float64
+    @test size(bsr_sim) == (4, 4)
+    @test typeof(bsr_sim) <: CuSparseMatrixBSR
+    @test bsr_sim.blockDim == bsr_mat.blockDim
```
maleadt: There's still these unrelated tests.

rainerrodrigues: Sorry about that, I have  edited that out as well! It should be fine now.

maleadt:
Thanks for the PR!

PR #3120, opened 2026-04-29
Bump PTX `.target` to match `--gpu-name`

AntonOresten:
I want to emit inline PTX through `@asmcall` on newer hardware. Suppose I want to use `add.f32x2`, gated on `sm_100`+, on a consumer Blackwell device (`sm_121`):

```julia
using CUDACore
using LLVM.Interop: @asmcall

function k()
    @asmcall("add.f32x2 \$0, \$1, \$2;", "=l,l,l", true,
               UInt64, Tuple{UInt64, UInt64}, UInt64(0), UInt64(0))
    return
end

@cuda k()
```

I get:

```
ERROR: Failed to compile PTX code (ptxas exited with code 255)
Invocation arguments: --generate-line-info --verbose --gpu-name sm_121 --output-file /tmp/jl_1WpIjMpMtY.cubin /tmp/jl_3AZwTS5w3o.ptx
ptxas /tmp/jl_3AZwTS5w3o.ptx, line 25; error   : Feature 'add.f32x2' requires .target sm_100 or higher
```

where the PTX generated by the outdated NVPTX from Julia 1.12's LLVM 18 emits `sm_90` as target:

```
shell> head /tmp/jl_3AZwTS5w3o.ptx
//
//

.version 9.2
.target sm_90
```

meaning that the assumption in the code ("no need to bump the `.target` directive; we can do that by passing `-arch` to `ptxas`") doesn't hold up.

The fix mirrors the existing `.version` rewrite for `.target`, in a small `rewrite_ptx_header` helper extracted from `mcgen`.

With these changes, one can call e.g. `add.f32x2`, which is one of only a few operations that are not `a`-gated, meaning it doesn't need `.target sm_100a` and `--gpu-name sm_100a`. Architecture- and family-specific ops will require another PR.

### DIFF
--- a/CUDACore/src/compiler/compilation.jl
+++ b/CUDACore/src/compiler/compilation.jl
@@ -122,7 +122,20 @@ function GPUCompiler.finish_module!(@nospecialize(job::CUDACompilerJob),
     return entry
 end
 
+# stamp `.version` with the ISA we want `ptxas` to validate against
+# and `.target` with the arch that `--gpu-name` will use
 function rewrite_ptx_header(asm, ptx, cap)
+    return replace(asm,
+        r"(\.version .+)"     => ".version $(ptx.major).$(ptx.minor)",
+        r"\.target sm_\d+\w*" => ".target sm_$(cap.major)$(cap.minor)")
+end
+
+function GPUCompiler.mcgen(@nospecialize(job::CUDACompilerJob), mod::LLVM.Module, format)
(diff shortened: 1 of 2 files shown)


maleadt:
> meaning that the assumption in the code ("no need to bump the `.target` directive; we can do that by passing `-arch` to `ptxas`") doesn't hold up.

Well, it does wrt. generating code. It only trips up `ptxas`' validation.
That said, I think this is pretty safe, so LGTM.

PR #3124, opened 2026-04-30
Add support for family- and architecture-specific features

AntonOresten:
Adds support for family-specific (`sm_NNNf`) and architecture-specific (`sm_NNNa`) PTX targets, enabling access to a wider set of low-level instructions. Builds on #3120; `PTXCompilerTarget` has no field for the suffix, so we rewrite `.target` and pass the matching `--gpu-name` to `ptxas`, sidestepping LLVM's NVPTX coverage.

NVIDIA defines three feature sets:

- baseline (no suffix, e.g. `sm_90`): forward-compatible.
- family (`f` suffix, e.g. `sm_100f`): same-major-family-portable. Requires CC ≥ 10.0 and PTX ≥ 8.8.
- architecture (`a` suffix, e.g. `sm_90a`): locked to one exact CC. Requires CC ≥ 9.0 and PTX ≥ 8.0.

with hierarchy baseline ⊆ family ⊆ architecture.

The previous behavior remains as the default through `:baseline`. To unblock `wgmma`, `tcgen05`, and friends, explicit opt-in is required through a new `feature_set` kwarg on `cufunction`, `@cuda`, etc. that takes a `Symbol`.

Relevant docs:
- [PTX ISA - Directives](https://docs.nvidia.com/cuda/parallel-thread-execution/#ptx-module-directives-target)
- [CUDA Programming Guide - Compute Capabilities](https://docs.nvidia.com/cuda/cuda-programming-guide/05-appendices/compute-capabilities.html#feature-availability)

Supercedes #3122 (renamed branch; seems beyond rescue)


AntonOresten:
While I am convinced `:baseline` as the default is correct for backward-compatibility (in CUDA.jl) and forward-compatiblity (in PTX?), `@cuda feature_set=:architecture` is quite verbose, especially considering `@cuda` only takes arguments on a single line. Perhaps it can be made to take blocks (see https://github.com/JuliaGPU/cuTile.jl/issues/40#issuecomment-3886210889), or shortening `:architecture` -> `:arch`.

EDIT: Shortened `:architecture` to `:arch`.

maleadt:
It's unfortunate that this splits the knowledge about architectures in multiple places; before it was only in the databases of `compatibility.jl`.

> Builds on #3120; `PTXCompilerTarget` has no field for the suffix, so we rewrite `.target` and pass the matching `--gpu-name` to `ptxas`, sidestepping LLVM's NVPTX coverage.

I'm not sure this is reasonable. NVPTX does actually have subtarget-specific emission paths, see e.g. `hasConvertWithStochasticRounding`: https://github.com/llvm/llvm-project/blob/477c59bb404af185976a23bc262dbeb09f788bff/llvm/lib/Target/NVPTX/NVPTXSubtarget.h#L305-L307

AntonOresten:
> NVPTX does actually have subtarget-specific emission paths

Right, I see your concern:

- LLVM is currently told e.g. `sm_103`, and emits PTX using only the baseline subtarget features. That PTX is, by superset, valid for `sm_103a` too.
- `.target` rewrite + `--gpu-name sm_103a`: tells ptxas "treat this stream as sm_103a". ptxas then accepts arch-specific instructions (but they can only appear via inline asm, since LLVM didn't emit any) and assembles for that exact CC.

So this PR unblocks the inline-PTX use case, but to get the LLVM-side benefits we'd want to thread the feature set into `PTXCompilerTarget` so the `TargetMachine` is constructed with `sm_103a`, which flips `hasArchAccelFeatures()` and lets the guarded patterns match.

I started at the CUDA.jl level because my GPUCompiler.jl mental model isn't deep enough yet to confidently extend `PTXCompilerTarget`.

Happy to scope this PR to the inline-PTX case and file a follow-up against GPUCompiler.jl for the `PTXCompilerTarget` change.

> It's unfortunate that this splits the knowledge about architectures in multiple places; before it was only in the databases of `compatibility.jl`.

The only suffix-aware code that ended up elsewhere is `format_target` (suffix -> `sm_NNN[a|f]` string) in `compilation.jl`. Could move that over too if it helps.

maleadt:
Here's the GPUCompiler.jl bit: https://github.com/JuliaGPU/GPUCompiler.jl/pull/798

maleadt:
Reworked significantly. On the back-end side things are now properly passed down to GPUCompiler.jl, but I also tried to make things more idiomatic on the front-end side. There's a `sm"..."` string-macro now to create version numbers with a feature set attached to them, and we should correctly select architecture-specific capabilities automatically (which includes handling a bunch of corner cases).

CUDACore/src/compiler/sm.jl
```diff
@@ -0,0 +1,73 @@
...
+
+    function SMVersion(major::Integer, minor::Integer, feature_set::Symbol = :baseline)
+        feature_set in (:baseline, :family, :arch) ||
+            error("SMVersion feature_set must be one of :baseline, :family, :arch; got $(repr(feature_set))")
+        return new(Int(major), Int(minor), feature_set)
+    end
+end
+
+function Base.parse(::Type{SMVersion}, s::AbstractString)
+    m = match(r"^(\d+)\.(\d+)([af]?)$", s)
```
AntonOresten: For SM versions specifically, I think it should be formatted without the dot, only permitting one minor digit. i.e.: `m = match(r"^(\d+)(\d)([af]?)$", s)`, which helps distinguish between compute capabilities and architecture.

CUDACore/src/compiler/sm.jl
```diff
@@ -0,0 +1,73 @@
...
+        return sm.major == dev_cap.major && base_version(sm) <= dev_cap
+    else  # :baseline
+        return base_version(sm) <= dev_cap
+    end
+end
+
+
+Base.show(io::IO, sm::SMVersion) = print(io, "sm\"", sm.major, ".", sm.minor, suffix(sm), "\"")
+
+macro sm_str(s)
```
AntonOresten: By no means necessary, but one could use the native string macro suffix feature to make the feature set a separate "argument", which would look like `sm"100"f`.

CUDACore/src/compiler/execution.jl
```diff
@@ -433,7 +433,11 @@ The following keyword arguments are supported:
 - `name`: override the name that the kernel will have in the generated code
 - `always_inline`: inline all function calls in the kernel
 - `fastmath`: use less precise square roots and flush denormals
-- `cap` and `ptx`: to override the compute capability and PTX version to compile for
+- `cap` and `ptx`: to override the compute capability and PTX version to compile for.
+  `cap` accepts an [`SMVersion`](@ref) via the `sm"..."` string macro, e.g.
```
AntonOresten: As mentioned in https://github.com/JuliaGPU/CUDA.jl/pull/3124/changes#r3266090477, I think it's worth distinguishing between capability and target, even though they're closely intertwined.

AntonOresten:
> and we should correctly select architecture-specific capabilities automatically

I initially chose to be conservative, but you're sure this won't compromise any downstream assumptions / caches if it becomes the default?

maleadt:
I don't think so.

CUDACore/src/compiler/sm.jl
maleadt: I guess. I don't see much value though? The suffix isn't really a modifier in the regex sense, I'd say it's more strongly part of the capability string.

AntonOresten: Right, it makes sense to have it in the SM version type instead of a hypothetical CC type, and the exact syntax doesn't matter. `sm"100f"` looks better,  even.

Just to clarify though, according to my understanding, the suffix is not part of the *capability*, which is a different category from the SM version: it houses the family- and arch-specific instruction sets, but sm_100a is not the capability. Rather, 10.0 is the capability that can run sm_100a targets. This is why I initially had `feature_set` as a separate keyword argument, trying to draw a line between them.

AntonOresten: For simplicity, it makes sense for `cap` to take `SMVersion`, but because of the distinction, it might more appropriately be named `target`. If passing `VersionNumber` is being deprecated anyway, I'm more inclined to deprecate `cap` altogether (if target is the only purpose it serves – not sure on this though).

maleadt: > Just to clarify though, according to my understanding, the suffix is not part of the _capability_, which is a different category compared to the SM version: it houses the family- and arch-specific instruction sets, but sm_100a is not the cap. Rather 10.0 is the cap that can run sm_100a targets. This is why I initially had `feature_set` as a separate keyword argument, trying to draw a line between them.

Yeah that sounds right. My mental model was code generation target vs device capabilities, and I guess some code paths now use SMVersion for both, or at least confound both concepts. On the other hand, it would be nice to use `sm"120"` for device capabilities too... I guess things would be more principled if the keyword argument were named `arch=` instead?

Let me try something out.

AntonOresten: That's better. Arguably the suffix isn't part of the architecture either though. The suffix just says which instructions of the architecture are allowed to be used.

maleadt: I realize, but I'm mimicking nvcc's `-arch=sm_120s` here. I don't think a more principled approach is worth splitting kwargs over?

AntonOresten: Oh, forgot about that. Sensible enough!

maleadt:
One other thing we need to consider is how to use this for conditional code paths (if we even want to support this conditionality at the host level). For example, `e5m2` etc are supported by `sm_120a`, and not the baseline architecture. Right now we'd be doing `if capability(device()) >= v"12.0" do ... end`, which doesn't extend to this design because the `a` suffix is a codegen modifier, not a device property.

You can query it from within the kernel, branching on `target_feature_set()`, but for type support we can't delay the decision that long.

AntonOresten:
> You can query it from within the kernel

So that'd be a device function like `CUDACore.compute_capability`?

> if we ever want to support this conditionality at the host level

If the user is branching on compute capability, they're controlling the kernel-launching? So they should be able to control the feature set as well. Since the feature set is now tied to the compute capability in the `cap` or `arch` argument, how would I change the target to e.g. family-specific? `arch=SMVersion(capability(device()), :family)`?

maleadt:
> > You can query it from within the kernel
> 
> So that'd be a device function like `CUDACore.compute_capability`?

Yes, and it's already in here (supported by the GPUCompiler.jl change).

> > if we ever want to support this conditionality at the host level
> 
> If the user is branching on compute capability, they're controlling the kernel-launching? So they should be able to control the feature set as well. Since the feature set is now tied to the compute capability in the `cap` or `arch` argument, how would I change the target to e.g. family-specific? `arch=SMVersion(capability(device()), :family)`?

Yes, I'm not thinking of people doing `@cuda`, but us implementing abstractions in CUDA.jl. That's the case where you want to write code that switches kernels based on what's possible to emit. I guess we could add an accessor for the current compiler config, which would return a SMVersion as opposed to a simple VersionNumber representing the capability.

CUDACore/src/compiler/sm.jl
AntonOresten: Sorry to deviate a bit from the original comment!😅 I believe we've resolved the thread's new scope though.

AntonOresten:
Are there any cases in which we don't want architecture-specific codegen, anyway? Linking my [earlier comment](https://github.com/JuliaGPU/CUDA.jl/pull/3124#issuecomment-4487583591)

maleadt:
When the toolchain doesn't support it?

maleadt:
Actually, I guess we can already do:
```julia
  if runs_on(sm"120a", capability(device()))
      @cuda arch=sm"120a" specialized_kernel(x)
  else
      @cuda fallback_kernel(x)
  end
```

So won't add another abstraction for now.

PR #3134, opened 2026-05-13
Improve runtime selecton

maleadt:
This PR bumps CUDA_Runtime_jll to a version that selects a runtime version based on the devices present in the system. Before, we only selected based on the driver version. We still consider the driver version, of course, but don't simply use the newest runtime anymore, instead taking into consideration whether that would rule out certain devices.


maleadt:
CI failure is unrelated.

But I'll repurpose this PR to test https://github.com/JuliaPackaging/Yggdrasil/pull/13712 once it merges. That change should make it possible to automatically select a compatible toolkit given the devices in the system.

PR #3138, opened 2026-05-14
Name register pressure in launch-failure diagnostics

AntonOresten:
## Problem

Register-pressure launch failures produce the generic kernel-thread-limit message, e.g.:

```
Number of threads per block exceeds kernel limit (258 > 256).
```

`256` here is the smallest of at least two numbers the driver minimizes over: the register-budget thread count (`floor(reglim / NUM_REGS)`, warp-aligned) and the hardware per-block cap (1024). The message doesn't say which is binding, so users have to query `FUNC_ATTRIBUTE_NUM_REGS` and back-solve to confirm registers were the cause.

## Fix

Insert a register-specific check before the kernel-thread-limit check, firing when `NUM_REGS * nthreads > MAX_REGISTERS_PER_BLOCK`:

```
Register pressure exceeds device limit (255 regs/thread * 258 threads/block = 65790 regs/block > 65536 regs/block).
```

This is strictly a refinement rather than a new catch: the new check's condition (`NUM_REGS * nthreads > reglim`) implies the existing threadlim condition (since `MAX_THREADS_PER_BLOCK ≤ floor(reglim / NUM_REGS)`), so detection is unchanged.

### DIFF
--- a/CUDACore/lib/cudadrv/execution.jl
+++ b/CUDACore/lib/cudadrv/execution.jl
@@ -161,7 +161,8 @@ end
     nregs = fattr[FUNC_ATTRIBUTE_NUM_REGS]
     reg_lim = attribute(dev, DEVICE_ATTRIBUTE_MAX_REGISTERS_PER_BLOCK)
     if nregs * nthreads > reg_lim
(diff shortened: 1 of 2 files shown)


maleadt:
I like this, but to a newcomer it may suggest having to lower register pressure, while lowering the block size would work as well. Maybe we should add suggestions ("Try lowering register pressure or launching fewer threads")?

maleadt (approved):

PR #3149, opened 2026-05-20
Reduce usage of `libdevice`, relying more on LLVM

maleadt:
Building on https://github.com/JuliaGPU/GPUCompiler.jl/pull/805, https://github.com/JuliaGPU/GPUCompiler.jl/pull/804, https://github.com/JuliaGPU/GPUCompiler.jl/pull/800, this PR aims to avoid some of the uses of `libdevice`'s intrinsics, instead emitting vanilla LLVM IR and having GPUCompiler.jl post-process it into what we need in PTX. This has many advantages, including (potentially) better optimization, compatibility with LLVM tools like Enzyme, etc.

cc @vchuravy


CUDACore/src/device/intrinsics/math.jl
```diff
@@ -508,27 +473,11 @@ end
...
-    inv_x = ccall("llvm.nvvm.rcp.approx.ftz.d", llvmcall, Float64, (Float64,), x)
-
-    # Approximate the missing 32bits of mantissa with a single cubic iteration
-    e = fma(inv_x, -x, 1.0)
-    e = fma(e, e, e)
-    inv_x = fma(e, inv_x, inv_x)
-end
+# `Base.FastMath.inv_fast(::AbstractFloat)` is unimplemented upstream (only
+# `Complex` has a method) and the catch-all fallback drops `afn`
+@device_override FastMath.inv_fast(x::Union{Float32, Float64}) =
```
vchuravy:
```suggestion
@device_override FastMath.inv_fast(x::Union{Float16, Float32, Float64}) =
```

CUDACore/src/device/intrinsics/math.jl
```diff
@@ -549,13 +498,9 @@ end
 @device_override Base.hypot(x::Float64, y::Float64) = ccall("extern __nv_hypot", llvmcall, Cdouble, (Cdouble, Cdouble), x, y)
 @device_override Base.hypot(x::Float32, y::Float32) = ccall("extern __nv_hypotf", llvmcall, Cfloat, (Cfloat, Cfloat), x, y)
 
-@device_override Base.fma(x::Float64, y::Float64, z::Float64) = ccall("llvm.fma.f64", llvmcall, Cdouble, (Cdouble, Cdouble, Cdouble), x, y, z)
-@device_override Base.fma(x::Float32, y::Float32, z::Float32) = ccall("llvm.fma.f32", llvmcall, Cfloat, (Cfloat, Cfloat, Cfloat), x, y, z)
-@device_override Base.fma(x::Float16, y::Float16, z::Float16) = ccall("llvm.fma.f16", llvmcall, Float16, (Float16, Float16, Float16), x, y, z)
-
-@device_override Base.muladd(x::Float64, y::Float64, z::Float64) = ccall("llvm.fmuladd.f64", llvmcall, Cdouble, (Cdouble, Cdouble, Cdouble), x, y, z)
-@device_override Base.muladd(x::Float32, y::Float32, z::Float32) = ccall("llvm.fmuladd.f32", llvmcall, Cfloat, (Cfloat, Cfloat, Cfloat), x, y, z)
-@device_override Base.muladd(x::Float16, y::Float16, z::Float16) = ccall("llvm.fmuladd.f16", llvmcall, Float16, (Float16, Float16, Float16), x, y, z)
```
vchuravy, line 558: That is something, Julia is currently not emitting. On the CPU it can sometimes improve vectorization to emit mul + add, but on the GPU we have seen it hurt performance.

test/core/codegen.jl
```diff
@@ -1,63 +1,47 @@
...
-        return
+@testset "assume" begin
+    @test @filecheck CUDA.code_llvm(Tuple{Int}) do i
+        @check "@gpu_report_exception"
+        cld(42, i)
     end
 
-    for (T, suffix) in ((Float32, "f32"), (Float64, "f64"), (Float16, "f16"))
-        ir = sprint(io->CUDA.code_llvm(io, muladd_kernel, Tuple{Ptr{T}}))
-        @test occursin("llvm.fmuladd.$suffix", ir)
```
vchuravy, line 46: We shoudl restore this

CUDACore/src/device/intrinsics/math.jl
maleadt: Should we do this using a fusion pass instead?

test/core/codegen.jl
maleadt: Well, I am checking this at the PTX level:
```
    @testset "fma/muladd PTX" begin
        # `Base.fma` lowers to `llvm.fma.fXX` (have_fma branch folded for
        # f32/f64 by GPUCompiler; for f16 we keep an explicit override).
        # `Base.muladd` lowers to `fmul contract + fadd contract`, which the
        # backend fuses. Either way: a single `fma.rn` per type.
        for (T, s) in ((Float16, "f16"), (Float32, "f32"), (Float64, "f64"))
            @test @filecheck CUDA.code_ptx(Tuple{T, T, T}) do x, y, z
                @check "fma.rn.$s"
                @check_not "__nv_fma"
                fma(x, y, z)
            end
            @test @filecheck CUDA.code_ptx(Tuple{T, T, T}) do x, y, z
                @check "fma.rn.$s"
                muladd(x, y, z)
            end
        end
    end
```

So aren't we properly contracting this during back-end optimization/compilation? Why is it bad to have it split at the LLVM level?

CUDACore/src/device/intrinsics/math.jl
```diff
@@ -286,44 +289,15 @@ end
...
 ## roots and powers
 
-@device_override Base.sqrt(x::Float64) = ccall("extern __nv_sqrt", llvmcall, Cdouble, (Cdouble,), x)
-@device_override Base.sqrt(x::Float32) = ccall("extern __nv_sqrtf", llvmcall, Cfloat, (Cfloat,), x)
-# sqrt(::Float16) inherits from Julia (Float16(sqrt(Float32(x)))), routing through __nv_sqrtf.
-@device_override FastMath.sqrt_fast(x::Union{Float32, Float64}) = sqrt(x)
-
 @device_function rsqrt(x::Float64) = ccall("extern __nv_rsqrt", llvmcall, Cdouble, (Cdouble,), x)
 @device_function rsqrt(x::Float32) = ccall("extern __nv_rsqrtf", llvmcall, Cfloat, (Cfloat,), x)
 @device_function rsqrt(x::Float16) = Float16(rsqrt(Float32(x)))
```
vchuravy: LLVM has a test like:

```
define float @rsqrt_fast(float %a) {
; CHECK-LABEL: rsqrt_fast(
; CHECK:       {
; CHECK-NEXT:    .reg .b32 %r<3>;
; CHECK-EMPTY:
; CHECK-NEXT:  // %bb.0:
; CHECK-NEXT:    ld.param.b32 %r1, [rsqrt_fast_param_0];
; CHECK-NEXT:    rsqrt.approx.f32 %r2, %r1;
; CHECK-NEXT:    st.param.b32 [func_retval0], %r2;
; CHECK-NEXT:    ret;
  %b = tail call afn float @llvm.sqrt.f32(float %a)
  %ret = fdiv afn float 1.0, %b
  ret float %ret
}
```

CUDACore/src/device/intrinsics/math.jl
```diff
@@ -508,27 +473,11 @@ end
...
-
-    # Approximate the missing 32bits of mantissa with a single cubic iteration
-    e = fma(inv_x, -x, 1.0)
-    e = fma(e, e, e)
-    inv_x = fma(e, inv_x, inv_x)
-end
+# `Base.FastMath.inv_fast(::AbstractFloat)` is unimplemented upstream (only
+# `Complex` has a method) and the catch-all fallback drops `afn`
+@device_override FastMath.inv_fast(x::Union{Float16, Float32, Float64}) =
+    FastMath.div_fast(one(x), x)
```
vchuravy, line 482: https://github.com/JuliaLang/julia/pull/61864

vchuravy (approved):

CUDACore/src/device/intrinsics/math.jl
```diff
@@ -549,13 +498,20 @@ end
...
+# `Base.fma(::Float16,...)` branches on `jl_have_fma`
+@device_override Base.fma(x::Float16, y::Float16, z::Float16) =
+    ccall("llvm.fma.f16", llvmcall, Float16, (Float16, Float16, Float16), x, y, z)
+
+# `Base.muladd(x, y, z) = fma(x, y, z)` is the natural choice on GPU: NVPTX
+# always lowers `llvm.fmuladd.fXX` to `fma.rn`, and routing through
+# `llvm.fmuladd` (rather than Julia's default `fmul contract + fadd contract`)
+# keeps the fusion robust under vectorization (per JuliaGPU/CUDA.jl#3149).
+@device_override Base.muladd(x::Float64, y::Float64, z::Float64) =
+    ccall("llvm.fmuladd.f64", llvmcall, Cdouble, (Cdouble, Cdouble, Cdouble), x, y, z)
```
gbaraldi, line 513: We should really do this upstream as well. Our mulladd could technically leak in the way we emit it.

CUDACore/src/device/intrinsics/math.jl
maleadt: Only on recent versions of LLVM, and the new FSqrt pass in GPUCompiler kills the pattern. So we need our own pattern if we want to write nice IR here; https://github.com/JuliaGPU/GPUCompiler.jl/pull/807.

That PR would allow us to write:
```julia
@device_function rsqrt(x::Float64) = @fastmath 1.0 / sqrt(x)
@device_function rsqrt(x::Float32) = @fastmath 1f0 / sqrt(x)
```

... which looks kinda funny because it uses `@fastmath` by default. But that is what `__nv_rsqrt` expanded to before, so I guess this is fine?

maleadt: Actually, I'm not sure that's fine. With the `@fastmath` we're telling LLVM the intrinsic can't ever return NaN, which isn't true.

maleadt: Let's just go with the NVVM intrinsic for now.

PR #3150, opened 2026-05-20
EnzymeCoreExt: overlay autodiff/deferred_codegen for kernel codegen

gbaraldi:
Companion to [EnzymeAD/Enzyme.jl#3112](https://github.com/EnzymeAD/Enzyme.jl/pull/3112) — addresses [EnzymeAD/Enzyme.jl#3091](https://github.com/EnzymeAD/Enzyme.jl/issues/3091).

## Summary

Two new overlays in \`CUDA.method_table\` inside \`ext/EnzymeCoreExt.jl\`:

1. **\`EnzymeCore.deferred_codegen(id::UInt) -> Ptr{Cvoid}\`** — the body holds the \`ccall(\"extern deferred_codegen\", llvmcall, …)\` marker that GPUCompiler's scanner picks up to link the inner Enzyme adjoint into the kernel module. EnzymeCore declares \`deferred_codegen\` as a method-less generic function so that \`jl_compile_all_defs\` can't despecialize the llvmcall into a host sysimage. CUDA's overlay supplies the body during kernel compilation; Enzyme.jl provides a matching overlay in \`GPUCompiler.GLOBAL_METHOD_TABLE\` for its own interpreter.

2. **\`EnzymeCore.autodiff(::ReverseMode|::ForwardMode, …)\` redirect** — inside CUDA's \`GPUInterpreter\` this forwards to \`autodiff_deferred\`, so user kernels can write plain \`Enzyme.autodiff(...)\` and get the deferred-codegen path automatically. Removes the manual \`autodiff\` vs \`autodiff_deferred\` choice that KernelAbstractions's \`gpu_fwd\` makes today.

The overlays only use EnzymeCore types, so no new weakdep is needed.

## Test plan

- [ ] CI green
- [ ] Verified locally with a CUDA kernel calling plain \`Enzyme.autodiff(Reverse, kernel, ...)\` — gradient comes out correct
- [ ] Existing \`test/extensions/enzyme.jl\` keeps passing

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -190,19 +190,26 @@ steps:
 
             println("--- :julia: Instantiating project")
             withenv("JULIA_PKG_PRECOMPILE_AUTO" => 0) do
+              # TODO(#3150): revert to registered Enzyme before merge. Points at
+              # EnzymeAD/Enzyme.jl#3112 for the matching `GLOBAL_METHOD_TABLE`
+              # overlay + `EnzymeCore._deferred_marker` shell.
+              enzyme_spec = PackageSpec(; name="Enzyme",
+                                          url="https://github.com/EnzymeAD/Enzyme.jl",
+                                          rev="deferred-codegen-overlay")
+
               # add Enzyme to the test deps
               Pkg.activate("test")
-              Pkg.add(["Enzyme", "EnzymeCore"])
+              Pkg.add([enzyme_spec, PackageSpec("EnzymeCore")])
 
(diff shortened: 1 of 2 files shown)


gbaraldi:
Closing in favor of the new approach in EnzymeAD/Enzyme.jl#3112: the fix for #3091 is now self-contained on the Enzyme side (weak-linkage llvmcall body) and doesn't need a CUDA-side overlay. The autodiff → autodiff_deferred redirect that was bundled here can be a separate PR if/when the existing cufunction-rule signature mismatch (Forward Kernel test failure on Julia 1.12) is resolved.

PR #3157, opened 2026-05-28
Fix pointer types in `cublasHgemm`

lpawela:
In the current main branch this fails
```julia
N = 256
A = CUDA.rand(Float16, N, N)
B = CUDA.rand(Float16, N, N)
C = CUDA.zeros(Float16, N, N)
CUDA.CUBLAS.gemm!('N', 'N', one(Float16), A, B, zero(Float16), C)
```
with
```julia
ERROR: MethodError: no method matching unsafe_convert(::Type{Ptr{Float16}}, ::Float16)
The function `unsafe_convert` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  unsafe_convert(::Type{Cwstring}, ::Any)
   @ Base strings/cstring.jl:101
  unsafe_convert(::Type{Ptr{T}}, ::Array{T}) where T
   @ Base pointer.jl:65
  unsafe_convert(::Type{Ptr{T}}, ::Ptr{NTuple{N, T}}) where {N, T}
   @ Base refpointer.jl:188

```
This PR fixes this issue.

### DIFF
--- a/lib/cublas/src/libcublas.jl
+++ b/lib/cublas/src/libcublas.jl
@@ -6199,9 +6199,9 @@ end
     initialize_context()
     @ccall libcublas.cublasHgemm(handle::cublasHandle_t, transa::cublasOperation_t,
                                  transb::cublasOperation_t, m::Cint, n::Cint, k::Cint,
-                                 alpha::Ptr{Float16}, A::Ptr{Float16}, lda::Cint,
-                                 B::Ptr{Float16}, ldb::Cint, beta::Ptr{Float16},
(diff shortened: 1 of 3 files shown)


maleadt (approved):
You cannot edit `libcublas.jl` directly, instead (or also) edit `res/wrap/libcublas_epilogue.jl` so that these changes don't get lost on the next toolkit release.

maleadt (changes requested):
You cannot edit `libcublas.jl` directly, instead (or also) edit `res/wrap/libcublas_epilogue.jl` so that these changes don't get lost on the next toolkit release.

lpawela:
@maleadt updated

PR #3162, opened 2026-06-04
Use an up-to-date LLVM by relying on an external `llc`.

maleadt:


maleadt:
Quiet some performance regressions here. Looking into them.

maleadt:
Looks like an LLVM regression wrt. small/under-aligned alloca's. Workaround in GPUCompiler:
```patch
diff --git a/src/ptx.jl b/src/ptx.jl
index 7c4ea5c..ffd178e 100644
--- a/src/ptx.jl
+++ b/src/ptx.jl
@@ -343,9 +343,49 @@ function finish_ir!(@nospecialize(job::CompilerJob{PTXCompilerTarget}),
         end
     end

+    raise_alloca_alignment!(mod)
+
     return entry
 end

+# raise the alignment of under-aligned allocas
+#
+# SelectionDAG normally raises the alignment of under-aligned stack objects when
+# expanding small memcpys (`DstAlignCanChange` in `getMemcpyLoadsAndStores`), but that
+# requires the destination to be a bare `FrameIndex` node. The NVPTX back-end wraps
+# allocas in `addrspacecast`s (NVPTXLowerAlloca), hiding the frame object, so the
+# expansion has to honor the IR-level alignment. That lowers e.g. the 7-byte padding
+# copies SROA generates for non-power-of-two aggregates to byte-wise loads and stores.
+# Compensate by raising the alignment of small allocas ourselves before codegen.
+function raise_alloca_alignment!(mod::LLVM.Module)
+    changed = false
+    @tracepoint "raise alloca alignment" begin
+
+    dl = datalayout(mod)
+    for f in functions(mod), bb in blocks(f), inst in instructions(bb)
+        inst isa LLVM.AllocaInst || continue
+
+        # only static allocas
+        array_size = operands(inst)[1]
+        array_size isa LLVM.ConstantInt || continue
+
+        alloca_type = LLVMType(LLVM.API.LLVMGetAllocatedType(inst))
+        size = abi_size(dl, alloca_type) * convert(Int, array_size)
+        size > 0 || continue
+
+        # match what `DstAlignCanChange` would have done: align to the largest
+        # power-of-two memory access this object could be copied with.
+        align = min(nextpow(2, size), 16)
+        if alignment(inst) < align
+            alignment!(inst, align)
+            changed = true
+        end
+    end
+
+    end
+    return changed
+end
+
 @unlocked function mcgen(@nospecialize(job::CompilerJob{PTXCompilerTarget}),
                          mod::LLVM.Module, format=LLVM.API.LLVMAssemblyFile)
     if !isavailable(NVPTX_LLVM_Backend_jll) || !NVPTX_LLVM_Backend_jll.is_available()
```

Let's first try to upstream this though: https://github.com/llvm/llvm-project/pull/201772

PR #3163, opened 2026-06-04
Add dp4a device intrinsic

JohnCobbler:
Adds `CUDACore.dp4a` — the packed 4-element int8/uint8 dot product with 32-bit accumulate (single PTX `dp4a` instruction, sm_61+), in its four signedness variants:

```julia
dp4a(a::Int32,  b::Int32,  c::Int32)  -> Int32   # dp4a.s32.s32
dp4a(a::Int32,  b::UInt32, c::Int32)  -> Int32   # dp4a.s32.u32
dp4a(a::UInt32, b::Int32,  c::Int32)  -> Int32   # dp4a.u32.s32
dp4a(a::UInt32, b::UInt32, c::UInt32) -> UInt32  # dp4a.u32.u32

```

Non-exported, `@device_function`, added to `@public` — following the `popc`/`byte_perm` precedent. Useful for quantized int8 inference and similar integer-heavy kernels.

## Implementation

Per the review discussion: on LLVM 21+ (which added `@llvm.nvvm.idp4a.[us].[us]`) the implementation uses the intrinsics via `ccall`; on older LLVM it falls back to inline PTX via `@asmcall` (same approach as `nanosleep`). Gated with `@static if LLVM.version() >= v"21"`, following the existing pattern in `math.jl`.

## Testing

- `test/core/device/intrinsics/math.jl`: GPU kernels for all four variants against a pure-Julia reference over edge cases (±127, -128, 255, all-ones, accumulator pass-through, mixed signs), plus a `@device_code_ptx` check pinning actual `dp4a` instruction selection (no emulation).
- Both paths verified on a Quadro RTX 6000 (sm_75): asm path on Julia 1.11.9 (LLVM 16), intrinsic path on nightly (LLVM 21.1.8). Confirmed via `@device_code_llvm` that each run took the intended path; identical `dp4a` instruction selection and identical results (4608 cases per path, zero mismatches).
- Instruction selection for the intrinsics additionally checked at the LLVM level: all four lower to their `dp4a.*` forms with the `NVPTX_LLVM_Backend_jll` llc (22.1.7).

Possible follow-up: `dp2a` (2-element int16×int8 dot product) — same pattern, separate PR if there's interest.


maleadt:
I guess that's one remaining limitation of #3162: intrinsics still have to exist in Julia's LLVM version. I guess we could relax `ccall`'s semantics, but using `@asmcall` is fine for now.

maleadt (approved):
Let's wait for CI to come back first.

vchuravy:
I would prefer a version check and using the LLVM intrinsic when they become available in the host compiler.

maleadt:
We could also do something funky like emitting `extern @llvm_new.foobar`, rewriting in GPUCompiler, but that seems sketchy.

test/core/device/intrinsics/math.jl
```diff
@@ -339,6 +339,134 @@ using SpecialFunctions
...
+                @cuda threads=1 kernel_uu(buf, a, b, c)
+                @test Array(buf)[] == expected
+            end
+        end
+
+        @testset "PTX instruction selection" begin
+            # Verify the backend emits the actual dp4a instruction, not a
+            # software emulation sequence.
+            buf = CuArray{Int32}(undef, 1)
+            ptx = sprint(io->(@device_code_ptx io=io @cuda launch=false kernel_ss(buf, Int32(0), Int32(0), Int32(0))))
```
maleadt: Please make this a `CUDA.code_ptx` or so, no need to actually launch kernels just to inspect generated code.

JohnCobbler:
switched PTX check to CUDA.code_ptx; verified locally

Issue #3165, opened 2026-06-05
cublasXt device-pointer tests intermittently read back all-NaN on multi-GPU CI


JohnCobbler:
On [cuda-dot-jl build 7447](https://buildkite.com/julialang/cuda-dot-jl/builds/7447) (MultiGPU, gpuci.16, 2× RTX 5000), every device-pointer cublasXt testset failed with the GPU result reading back as all-NaN:

```
Error in testset xt_trmm! gpu:
  Expression: C ≈ h_C
  Evaluated: Float32[0.3404317 0.31222013 …] ≈ Float32[NaN NaN …]   # h_C entirely NaN
```

All host-pointer (`cpu`) variants passed in the same run, and the same agent was green on builds 7440–7446, so this looks like a rare multi-GPU race rather than anything in the PR under test (#3163; its own testsets passed).

It smells like a relative of #2320: `xt_handle()` grants pool access to the other devices with `access!` but doesn't check `can_access_peer` first (unlike `maybe_enable_peer_access` in CUDACore). The docs note unchecked `cuMemPoolSetAccess` on a fresh pool can succeed on non-peer-capable devices and only fail later — though that path should fail the next allocation loudly rather than produce NaN, so this may not be the whole story.

I can't reproduce locally (single GPU here). Happy to PR the defensive `can_access_peer` check if useful — it wouldn't explain everything, but it would turn one silent path loud. Full job log parsed evidence available if helpful.

PR #3166, opened 2026-06-05
cuBLASXt: only grant pool access to peer-capable devices

JohnCobbler:
As offered in #3165: `xt_handle()` grants pool access to all participating devices without checking `can_access_peer`, unlike `maybe_enable_peer_access` in CUDACore. Per the CUDA docs, an unchecked `cuMemPoolSetAccess` on a fresh pool can succeed on non-peer-capable devices, deferring the failure to a later allocation.

This checks peer capability first, warns (once) about devices that can't access a pool, and only grants access where valid.

Caveat up front: I only have a single GPU, so this is untested on actual multi-GPU hardware, and per the issue it likely isn't the whole story behind the NaN readbacks. Feel free to close if you'd rather investigate the root cause first.

### DIFF
--- a/lib/cublas/src/cuBLAS.jl
+++ b/lib/cublas/src/cuBLAS.jl
@@ -215,8 +215,16 @@ function xt_handle()
         async_devs = filter(memory_pools_supported, devices())
         for dev in async_devs
             other_devs = filter(!isequal(dev), async_devs)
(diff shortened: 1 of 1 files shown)


maleadt (approved):
LGTM, thanks.

PR #3173, opened 2026-06-10
[Backport 6.1]: Make the compiler params extensible

wsmoses:


maleadt:
I'm not going to release a backport for 6.1 (which is a whole hassle now with the staggered release, and the existing retrocaps in General for 6.1.x). 6.2 should be out later this week.

PR #3175, opened 2026-06-12
Update CI

maleadt:


kshyatt:
Is this what all CUDA CI users should be doing? I can start on the repos I have access to if so

maleadt:
I was going to script it, so no need.

kshyatt:
Too late!!

PR #3180, opened 2026-06-21
Add narrow-precision floats to cudaDataType

AntonOresten:
Adds narrow-precision floats to the cudaDataType enum based on `library_types.h` in CUDA 13:

```c
    CUDA_R_8F_E4M3 = 28, /* real as a nv_fp8_e4m3 */
    CUDA_R_8F_UE4M3 = CUDA_R_8F_E4M3, /* real as an unsigned nv_fp8_e4m3 */
    CUDA_R_8F_E5M2 = 29, /* real as a nv_fp8_e5m2 */
    CUDA_R_8F_UE8M0 = 30,  /* real as an exponent-only unsigned nv_fp8_e8m0 */
    CUDA_R_6F_E2M3  = 31,  /* real as a nv_fp6_e2m3 */
    CUDA_R_6F_E3M2  = 32,  /* real as a nv_fp6_e3m2 */
    CUDA_R_4F_E2M1  = 33,  /* real as a nv_fp4_e2m1 */
```

`R_8F_UE4M3` is an alias of `R_8F_E4M3`, differing semantically in that the sign is meaningless in the NVFP4 block-scaling format where every element has a sign anyway.

Implementations like [DLFP8Types.jl](https://github.com/chengchingwen/DLFP8Types.jl) and [Microfloats.jl](https://github.com/MurrellGroup/Microfloats.jl) could define e.g. `Base.convert(::Type{CUDACore.cudaDataType}, ::Type{Float8_E5M2}) = CUDACore.R_8F_E5M2` in an extension, or there could be some `jltype_to_cudaDataType` function.

It doesn't work the other way though, since each cudaDataType can only map to one julia type, but not sure when this would be needed.

### DIFF
--- a/CUDACore/src/library_types.jl
+++ b/CUDACore/src/library_types.jl
@@ -33,6 +33,12 @@ end
     C_64I  = 25 # complex as a pair of signed 64-bit int numbers
     R_64U  = 26 # real as a unsigned 64-bit int
     C_64U  = 27 # complex as a pair of unsigned 64-bit int numbers
+    R_8F_E4M3  = 28 # real as a nv_fp8_e4m3
+    R_8F_E5M2  = 29 # real as a nv_fp8_e5m2
+    R_8F_UE8M0 = 30 # real as an exponent-only unsigned nv_fp8_e8m0
+    R_6F_E2M3  = 31 # real as a nv_fp6_e2m3
+    R_6F_E3M2  = 32 # real as a nv_fp6_e3m2
+    R_4F_E2M1  = 33 # real as a nv_fp4_e2m1
 end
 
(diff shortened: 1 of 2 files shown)


AntonOresten:
@maleadt How, and where, could packages define `DataType` to `cudaDataType` mappings? Is this considered public API? If not, would it need to be in extensions of CUDACore?

maleadt:
> How, and where, could packages define `DataType` to `cudaDataType` mappings? Is this considered public API? If not, would it need to be in extensions of CUDACore?

It's semi-public; we advertise supporting the C API, which `cudaDataType` is, but there's no clean extension point. You can do `Base.convert(::Type{CUDACore.cudaDataType}, ::Type{MyType})` but can't extend the reverse.

PR #3186, opened 2026-07-03
don't promote to i64 unnecessarily in `unsafe_cached_load`

simeonschaub:
Might be more of a micro optimization, since eventually we will promote to i64 in `add_ptr` anyways, but it doesn't hurt


simeonschaub:
1.11 is failing with a resolver error and nightly failures look unrelated as well

PR #3188, opened 2026-07-05
Fix cluster barriers silently trapping on Julia <= 1.11

AntonOresten:
The cluster barriers use `ccall("llvm.nvvm.barrier.cluster.*", llvmcall, ...)`, which resolves the name against the loaded LLVM. These intrinsics only exist in LLVM 17+, so on Julia 1.10/1.11 the call lowers to a deferred runtime error: the kernel compiles cleanly and traps at launch, with no diagnostic. CI never sees it because the cluster tests require sm_90 hardware, which never runs with Julia 1.11.

Fix: emit them via the full-module `llvmcall` form (as `version.jl` does), so the name passes through to the NVPTX back end, with `convergent nomerge` spelled out since the loaded LLVM can't infer attributes for unknown names (`nomerge` because SimplifyCFG only learned to not merge convergent calls in LLVM 17).

Also adds a codegen test outside the capability gate — device IR must contain both calls and no `gpu_report_exception` — so it runs on all CI, including CPU-only. Verified on 1.10/1.11/1.12: IR and sm_90 PTX contain the barriers, and the new test fails on 1.11 without the fix.

Diagnosed by Fable 5

### DIFF
--- a/CUDACore/src/device/intrinsics/synchronization.jl
+++ b/CUDACore/src/device/intrinsics/synchronization.jl
@@ -72,11 +72,26 @@ export barrier_sync
 
 export cluster_arrive, cluster_arrive_relaxed, cluster_wait
 
-@device_functions begin
-cluster_arrive() = ccall("llvm.nvvm.barrier.cluster.arrive", llvmcall, Cvoid, ())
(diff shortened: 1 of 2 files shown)


maleadt:
1.11 CI failure needs a JLD2 version bump, I think...

PR #3192, opened 2026-07-10
Restore testing of libraries + fix regressions

maleadt:
Got lost during the switch to ParallelTestRunner...


maleadt:
Fixed, and reworked handling of ordinary errors too so that we can better report on unconditional exceptions instead of asserting (and bricking the process).

PR #3193, opened 2026-07-12
Remove some allocations from unsafe_wrap(::Type{CuArray{T,N,M}}, p::P…

ameligrana:
…tr{T}, dims::NTuple{N,Int})

This is better e.g.

```julia
julia> cpu = [1,2,3,4,5,6,7,8,9,10,11,12,13];

julia> @time gpu = unsafe_wrap(CuArray, cpu);
  0.000018 seconds (13 allocations: 624 bytes) # this pr
  0.000027 seconds (15 allocations: 688 bytes) # main
```

PR #3194, opened 2026-07-12
Implement unsafe_wrap! for in-place re-wrapping of non-owning CuArrays

ameligrana:
I'd like to wrap resizable CPU buffers as unified-memory CuArrays every time they are used. `unsafe_wrap` was problematic for me since it allocates: each call builds a new CuArray + DataRef + RefCounted + Managed and registers a finalizer which require 7 allocations on my machine (and some hundreds of ns). I was trying to improve the function in #3193 (which can still be useful but needs to be improved), but then I realized that maybe a function which operates in place could be better for me to implement everything I need.

I propose to add `unsafe_wrap!` which updates a non-owning `CuArray` (e.g. created by `unsafe_wrap`) to wrap the given memory and dims, reusing the array object. It is non-allocating and cheap to call before every use and so it covers my use case.

Hope this can be merged! It would be very useful for my use case to have something like this (or well, a better version if this is not correct in some way :-))

disclaimer: the changes were implemented with the help of Fable 5.


ameligrana:
ok, I actually found another way to do this which doesn't require this function. So maybe this is overkill actually for my use case...in any case maybe this can still be useful to someone else, so I keep this open

PR #3196, opened 2026-07-13
Make kernel launches allocation-free (on Julia 1.12+)

maleadt:
Use stack-promotable `Ref` storage for kernel arguments, launch configuration, attributes, and driver query outputs. Also use `cuStreamIsCapturing` for the common capture check.

On Julia 1.12, a warmed `@cuda launch` with real `CuArray` arguments now allocates zero bytes.

Before:
```julia-repl
julia> using CUDA, BenchmarkTools

julia> include("examples/vadd.jl")
Test Passed

julia> function launch_vadd!(a, b, c, len, count)
           for _ in 1:count
               @cuda threads=len vadd(a, b, c)
           end
           return
       end
launch_vadd! (generic function with 1 method)

julia> launch_vadd!(d_a, d_b, d_c, len, 100)

julia> CUDA.synchronize()

julia> @benchmark launch_vadd!($d_a, $d_b, $d_c, $len, 100) teardown=(CUDA.synchronize()) evals=1 samples=1_000
BenchmarkTools.Trial: 1000 samples with 1 evaluation per sample.
 Range (min … max):  156.804 μs … 228.017 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     163.156 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   171.002 μs ±  14.748 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

   ▄█▃▂                                                   ▆
  ▅████▇▆▆▄▆▅▄▄▃▄▅▃▃▂▃▂▂▂▂▃▃▂▃▃▃▄▃▃▂▂▃▃▃▂▂▃▂▂▂▂▂▂▂▁▁▂▁▂▂▃▇█▆▅▄▂ ▃
  157 μs           Histogram: frequency by time          197 μs <

 Memory estimate: 42.19 KiB, allocs estimate: 700.
```

After:

```julia-repl
julia> @benchmark launch_vadd!($d_a, $d_b, $d_c, $len, 100) teardown=(CUDA.synchronize()) evals=1 samples=1_000
BenchmarkTools.Trial: 1000 samples with 1 evaluation per sample.
 Range (min … max):  155.842 μs … 225.292 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     159.079 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   169.735 μs ±  15.671 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▄█▇▄▄▂          ▁▁                                     ▅▆▃▃
  ███████▇█▆▇▆██▇███▇▆▅▅▆▅▆▅▄▇▇▆▇██▇▇▇▄▆▅▇▇▆▆▆▅▁▄▆▄▅▄▅▅▄▁█████▄ █
  156 μs        Histogram: log(frequency) by time        197 μs <

```

cc @vchuravy

### DIFF
--- a/CUDACore/lib/cudadrv/execution.jl
+++ b/CUDACore/lib/cudadrv/execution.jl
@@ -36,37 +36,40 @@ function launch(f::CuFunction, args::Vararg{Any,N}; blocks::CuDim=1, threads::Cu
     threaddim = CuDim3(threads)
     clusterdim = CuDim3(clustersize)
 
-    attrs = CUDACore.CUlaunchAttribute[]
-    if cooperative
-        resize!(attrs, length(attrs)+1)
-        attr = pointer(attrs, length(attrs))
-        attr.id = CUDACore.CU_LAUNCH_ATTRIBUTE_COOPERATIVE
-        attr.value.cooperative = 1;
-    end
-    if clusterdim.x != 1 || clusterdim.y != 1 || clusterdim.z != 1
-        resize!(attrs, length(attrs)+1)
-        attr = pointer(attrs, length(attrs))
-        attr.id = CUDACore.CU_LAUNCH_ATTRIBUTE_CLUSTER_DIMENSION
-        attr.value.clusterDim.x = clusterdim.x
-        attr.value.clusterDim.y = clusterdim.y
-        attr.value.clusterDim.z = clusterdim.z
-    end
+    attributes = Ref{NTuple{2,CUlaunchAttribute}}()
+    GC.@preserve attributes stream begin
+        attributes_ptr = Base.unsafe_convert(Ptr{CUlaunchAttribute}, attributes)
+        num_attributes = 0
+        if cooperative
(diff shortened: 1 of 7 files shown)


CUDACore/lib/cudadrv/execution.jl
```diff
@@ -1,46 +1,16 @@
 # Execution control
 
-# In contrast to `Base.RefValue` we just need a container for both pass-by-ref (Symbol),
-# and pass-by-value (immutable structs).
-mutable struct ArgBox{T}
-    const val::T
```
vchuravy, line 6: Why the removal of ArgBox? We don't gain anything w.r.t. to Ref, but we lose  the semantic simplification

maleadt: The less code the better? But if you feel the simplification is worth it I'll put it back.

PR #3199, opened 2026-07-14
Fix 3-arg QR `ldiv!` for multi-column right-hand sides

hexaeder:
`ldiv!(x, ::QR, b)` in `lib/cusolver/src/linalg.jl` assumed a vector rhs:

```julia
_x = ldiv!(_qr, b)
x .= vec(_x)          # <- flattens an n×k result
```

The 2-arg `ldiv!(::QR, ::CuMatrix)` right above it already handles matrix rhs, but
the 3-arg form never got the same treatment. With a k>1 column rhs, `_x` is n×k and
`vec(_x)` is length n·k, so broadcasting into the n×k destination `x` throws
`DimensionMismatch: array could not be broadcast to match destination`.

`lib/cusolver/test/dense/qr.jl` only exercised `ldiv!` with vector rhs. Added
matrix-rhs cases for the 2-arg and 3-arg forms. The 3-arg one fails on master for
all four eltypes with the `DimensionMismatch` above; both pass with the fix.

### DIFF
--- a/lib/cusolver/src/linalg.jl
+++ b/lib/cusolver/src/linalg.jl
@@ -298,7 +298,7 @@ end
 
 function LinearAlgebra.ldiv!(x::CuArray, _qr::QR, b::CuArray)
     _x = ldiv!(_qr, b)
-    x .= vec(_x)
+    copyto!(x, _x)
     unsafe_free!(_x)
     return x
 end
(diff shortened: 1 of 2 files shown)


kshyatt (approved):
Nice, thanks for correcting this!

kshyatt:
Let's just rerun on current CI to make sure everything looks good then I'm happy to merge

kshyatt:
OK everything here looks ok to me, but the benchmarks aren't publishing? @maleadt is there something wrong with the benchmark pipeline?

kshyatt:
OK I'm going to try to merge this, 🤞

PR #3221, opened 2026-08-04
Delete unnecessary `axes()` method

JamesWrigley:
It already has a Base definition implemented in terms of `size()`.

Fixes these invalidations on 1.13:
```julia
 inserting axes(C::CUDACore.Const) @ CUDACore ~/.julia/packages/CUDACore/NlVPI/src/device/array.jl:216 invalidated:                                                                                                                            
   backedges: 1: superseding axes(A) @ Base abstractarray.jl:100 with MethodInstance for axes(::DenseVector{UInt8}) (180 children)
```


JamesWrigley:
(bump)

maleadt (approved):

PR #3241, opened 2026-08-18
cuDNN: accept F8_128x4 scale tiles on any blocked dimension

AntonOresten:
In #3230, `check_swizzled_scale_dims` validated F8_128x4 swizzled scale tensors positionally: innermost scale dimension padded to a multiple of 4, its partner to a multiple of 128. That happens to describe every ordinary matmul operand, where the blocked contraction dim is storage-innermost. The blocked axis's scale count pads to 4 and the partner axis to 128, whatever the storage order. The two readings diverge once an operand is blocked along a non-innermost axis, which is exactly what the MXFP8 attention recipe does to V (block-scaled along the sequence axis, head dim innermost). [cudnn-frontend's mxfp8 sample](https://github.com/NVIDIA/cudnn-frontend/blob/c2f4357/samples/cpp/sdpa/mxfp8_fwd.cpp#L112-L123) declares SF_V as `{b, h, round4(⌈s/32⌉), round128(d)}` and notes the declared inner stride "is not load-bearing". The kernel reads scales through the F8_128x4 reordering, so the old check would have rejected NVIDIA's own shapes.

- `check_swizzled_scale_dims` now tries (blocked, partner) dimension pairs; scale strides are no longer consulted.
- Unit extents participate: a `(1, k)` vector operand's partner axis has extent 1 and still pads to a whole 128-tile. Matching is exact equality against the declared dims, so admitting unit-extent pairs cannot false-accept — it only stops false-rejecting vector operands (a transposed NVFP4 gemv operand is the practical case).
- Whole-tile padding stays enforced: engines address whole 128×4 tiles regardless of declared extents, and unpadded declarations build fine but crash the context at execution (measured).

Tested with construction accept/reject cases for the frontend sample's V shape and the vector orientation, plus an execution test gated on engine availability: a pair-dequant matmul with the B operand's blocked axis not innermost, crafted E4M3/E8M0 bytes bound through the `checked_array_pointer` seam, checked against an exact reference. Uniform scales are invariant under the swizzle, so the bytes bind without reproducing the tile layout, and doubling both operands' scales must exactly quadruple the product — it does, to the bit, on sm_121 (GB10, cuDNN 9.24), where engines exist for both storage orientations.

Made with Claude Code

If this could be squeezed in before v6.3 I would be happy, as it's the last fix of my recent spree, but since it is a minor fix that only enables MXFP8 attention (which I have yet to test due to B200 availability), it's not too urgent. 😊

[only subpackages]

### DIFF
--- a/lib/cudnn/src/graph/ops.jl
+++ b/lib/cudnn/src/graph/ops.jl
@@ -846,17 +846,24 @@ function swizzled_scale_dims(xdims, block_size, order)
 end
 
 # engines address whole tiles regardless of the declared extents; cuDNN accepts
-# unpadded declarations at build time but crashes the context at execution
+# unpadded declarations at build time but crashes the context at execution.
+# Tile roles follow the quantization semantics, not storage order: the blocked
+# axis's scale count pads to a multiple of 4, its partner axis to 128.
 function check_swizzled_scale_dims(fname, x::Tensor, scale::Tensor, block_size)
-    order = [i for i in sortperm(scale.strides) if scale.dims[i] != 1]
-    expected = swizzled_scale_dims(x.dims, block_size, order)
-    scale.dims == expected || throw(DimensionMismatch(
+    round_up(v, m) = cld(v, m) * m
+    for p in eachindex(x.dims), q in eachindex(x.dims)
+        p == q && continue
+        expected = collect(Int64, x.dims)
+        expected[p] = round_up(cld(expected[p], block_size), 4)
+        expected[q] = round_up(expected[q], 128)
+        scale.dims == expected && return nothing
+    end
+    throw(DimensionMismatch(
         "$fname F8_128x4 scale dimensions $(Tuple(scale.dims)) do not tile data " *
-        "dimensions $(Tuple(x.dims)) with block_size=$block_size; the swizzled layout " *
-        "needs whole 128×4 tiles: cld(extent, block_size) scales rounded up to a " *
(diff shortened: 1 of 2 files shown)


maleadt:
> If this could be squeezed in before v6.3 I would be happy, as it's the last fix of my recent spree, but since it is a minor fix that only enables MXFP8 attention (which I have yet to test due to B200 availability), it's not too urgent. 😊

Sure, I'll wait until this is green. But yeah I'm hoping to release 6.3 ASAP, so if you want to do any last-minute validation of the new cuDNN APIs (thanks, by the way) now is definitely the time.

AntonOresten:
My previous PRs were all testable on consumer Blackwell; this is the first whose real consumer (MXFP8 attention) is CC 10.x-only. I managed to grab a cloud B200 just now, and numerically verified MXFP8 attention forward through the composed graph this check enables! 😁 While I had the box I also ran the full cuDNN test suite on CC 10.0 — 862 pass / 2 broken, zero failures, which also gives #3238's composite SDPA backward its first data-center Blackwell execution.

image

Btw, Is my `[only subpackages]` flag not being recognized? I've been putting it in the commit message, and this time I also put it in the description.

maleadt:
> Btw, Is my `[only subpackages]` flag not being recognized? I've been putting it in the commit message, and this time I also put it in the description.

It sometimes doesn't takes, I'm not sure why.

PR #3219, opened 2026-07-28
don't return reshaped array from `mapreducedim!`

simeonschaub:
Discovered in https://github.com/JuliaGPU/GPUArrays.jl/pull/754. Once this is approved, I will open PRs for all the other backends as well.

PR #3244, opened 2026-08-21
[release-6.2] Backports for 6.2

kshyatt:
`AbstractMatrix{T}` in the copied operand made these wider than GPUArrays' generic triangular solves in one slot and narrower in the others, so with JuliaGPU/GPUArrays.jl#763 dispatch is ambiguous.

(cherry picked from commit afa2910940a3b73bcec6a4865b49787978a9224b)

### DIFF
--- a/CUDACore/ext/EnzymeCoreExt.jl
+++ b/CUDACore/ext/EnzymeCoreExt.jl
@@ -37,6 +37,13 @@ function metaf(config, fn, args::Vararg{Any, N}) where N
 end
 
 function EnzymeCore.EnzymeRules.forward(config, ofn::Const{typeof(cufunction)},
+                                        ::Type{<:Const}, f::Const{F},
+                                        tt::Const{TT}; kwargs...) where {F,TT}
+    res = ofn.val(f.val, tt.val; kwargs...)
+    return res
+end
+
+function EnzymeCore.EnzymeRules.forward(config, ofn::Const{typeof(cufunction)},
                                         ::Type{<:Duplicated}, f::Const{F},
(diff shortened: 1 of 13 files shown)


maleadt (approved):

kshyatt:
Apparently things are no bueno with `mapreducedim!` though, I'll look into why

christiangnrd:
Needs https://github.com/JuliaGPU/CUDA.jl/pull/3219. I'm running tests locally and will push here if passing

kshyatt:
Could we pull in https://github.com/JuliaGPU/CUDA.jl/commit/58ed564254dfef476a81a8950c4b7b1145166194 and https://github.com/JuliaGPU/CUDA.jl/commit/8fc49feda7cf563fa9fb4900e329dfecc99ed0fd if they're not already on here?

christiangnrd:
Might as well they were clean merges

kshyatt:
I ask because this will allow people to successfully use Enzyme with CUDA while Enzyme itself updates to GPUCompiler 2.x

maleadt:
If we're doing backports anyway, feel free to put any other bug fixes on here, as long as CI passes.

kshyatt:
Are we good to merge if this passes CI?

christiangnrd:
The only other PRs I would maybe add are #3227 and #3198 but I'm not sure if they're too disruptive. They do pass locally but they seem like fixes for weirder configurations which my PC isn't.

kshyatt:
We could always follow this up with another patch release with just those two?

christiangnrd:
Good idea

PR #3246, opened 2026-08-22
KernelInterface

christiangnrd:
Do not merge until KernelInterface has been reviewed and interface fully decided


christiangnrd:
I'm pretty sure the record_event and wait_event are implemented properly, but Fable says the test doesn't always fail when wait_event is disabled because an array on a different stream gets automatically synchronized

CUDACore/src/CUDACore.jl
```diff
@@ -124,10 +125,14 @@ include("complex.jl")
 include("library_types.jl")
 
 # KernelAbstractions
-include("CUDAKernels.jl")
+include("CUDAKernelsOld.jl")
```
maleadt: This is ugly?

CUDACore/src/CUDAKernels.jl
```diff
@@ -58,185 +57,196 @@ function KA.copyto!(::CUDABackend, A, B)
...
+#     if blocks == 0
+#         return nothing
+#     end
+
+#     # Launch kernel
+#     CUDACore.kernel_launch(kernel, call; threads, blocks)
+
+#     return nothing
+# end
+
```
maleadt: Why the commented out stuff?

christiangnrd: I probably had it commented while moving things over and forgot to remove. It's gone now

CUDACore/src/CUDACore.jl
christiangnrd: Yes. I found it was much easier in terms of rebasing to have the KI code in its future spot right away and then just turn the new file into the extension in the KA PR. I'm open to suggestions

PR #3247, opened 2026-08-25
Extend Enzyme rules for SVectors

jgreener64:
The Enzyme rules are currently written for arrays of floats or complex floats, but other isbits types like SVectors are also valid in CuArrays. This is an attempt to broaden the rules to work there.

Since Enzyme depends on GPUCompiler 1.6.2 and CUDACore on GPUCompiler 2.2, this might require some other changes before it gets run on CI.

Tagging @wsmoses for review. Initially coded by me, refined by Claude.


wsmoses (approved):
lgtm

wsmoses:
note @jgreener64 @kshyatt this needs to be backported to the earlier cuda release to be usable

PR #3253, opened 2026-08-29
[cuSPARSE] fix Float16 spmm

simeonschaub:
Float16 spmm must use Float32 as the accumulation type, anything else is rejected by cuSPARSE. We were already using the correct accumulation type for spmv, this adds the same logic to spmm


maleadt:
cuSPARSE CI failures look related?

simeonschaub:
Looks like CUDA 13.1 and 13.2 don't support this, so skipped the tests on those configs

lib/cusparse/test/bmm.jl
```diff
@@ -244,3 +244,38 @@ end
         @test D ≈ C
     end
 end
+
+# Float16 must use Float32 as accumulation type, which is the only configuration
+# supported by cuSPARSE. The cuSPARSE versions shipped with CUDA 13.1 and 13.2
+# reject this configuration with CUSPARSE_STATUS_NOT_SUPPORTED, so skip it there.
+if !(Base.thisminor(CUDACore.runtime_version()) in (v"13.1", v"13.2"))
```
maleadt: CUSPARSE tests should gate on the CUSPARSE version, not the CTK one.

lib/cusparse/test/generic.jl
```diff
@@ -129,6 +129,34 @@ for SparseMatrixType in keys(SPMM_ALGOS)
     end
 end
 
+# Float16 must use Float32 as accumulation type, which is the only configuration
+# supported by cuSPARSE (same for ComplexF16). The cuSPARSE versions shipped with
+# CUDA 13.1 and 13.2 reject this configuration with CUSPARSE_STATUS_NOT_SUPPORTED,
+# so skip it there.
+if !(Base.thisminor(CUDACore.runtime_version()) in (v"13.1", v"13.2"))
```
maleadt, line 136: I think this still works fine under these versions; only batched mm! has issues.

PR #3261, opened 2026-09-04
Separate PTX and cubin target compatibility

maleadt:
CUDA.jl loads the cubin produced by `ptxas`, so PTX forward compatibility alone does not make a compilation target executable. For example, PTX targeting `sm_90` can be inspected on an `sm_120` GPU, but its `sm_90` cubin cannot load there.

Model PTX and cubin compatibility separately. Prefer the most capable target whose cubin can load, falling back to a PTX-compatible target for reflection. Kernel construction rejects incompatible cubins before compilation, including explicit `arch=` requests. Family and architecture targets retain their narrower compatibility rules, including Thor's `sm_101`/`sm_110` naming change. LLVM can still fall back to a supported baseline target when it lacks the requested variant.

Test target compatibility and selection independently of the installed GPU. Intrinsic requirement tests compile and validate without loading a cubin, so cross-generation targets reach the intended static assertions. The PTX reflection test only excludes architecture/family targets when using the cross-generation fallback.

### DIFF
--- a/CUDACore/src/compatibility.jl
+++ b/CUDACore/src/compatibility.jl
@@ -8,10 +8,11 @@ const highest = v"999"
 # selected via the suffix on the `.target` directive (and the matching `--gpu-name`
 # to ptxas):
 #
-#   - Baseline (no suffix, e.g. sm_90): the forward-compatible feature set. Code compiled
-#     for sm_X runs on any sm_Y with Y >= X (onion model).
-#   - Family (`f` suffix, e.g. sm_100f): a superset of Baseline. Same-major-family-portable;
-#     code compiled for sm_100f runs on sm_100, sm_103, etc., but not across families.
+#   - Baseline (no suffix, e.g. sm_90): the forward-compatible feature set. PTX compiled
+#     for sm_X runs on any sm_Y with Y >= X (onion model); cubins only within a major.
+#   - Family (`f` suffix, e.g. sm_100f): a superset of Baseline. Forward-compatible with
+#     later targets in the same architecture family, i.e. the same major version (with
+#     sm_101f belonging to the sm_11x family).
 #   - Architecture (`a` suffix, e.g. sm_90a): a superset of Family. Locked to one
 #     exact CC; code compiled for sm_103a runs only on CC 10.3 devices.
 #
--- a/CUDACore/src/compiler/compilation.jl
+++ b/CUDACore/src/compiler/compilation.jl
@@ -239,6 +239,34 @@ function default_ptx_versions(llvm_support, ptxas_support)
     return maximum(common_ptxs), maximum(ptxas_ptxs)
 end
 
+# Prefer a target whose cubin can execute on the device. If the toolchain cannot produce
+# one, retain the best PTX-compatible target for PTX-level reflection; execution rejects it.
+function select_ptxas_sm(ptx_sms, dev_cap)
+    supported = filter(sm -> base_version(sm) >= base_version(minreq.sm), ptx_sms)
+    candidates = filter(sm -> cubin_compatible(sm, dev_cap), supported)
+    if isempty(candidates)
+        candidates = filter(sm -> ptx_compatible(sm, dev_cap), supported)
+    end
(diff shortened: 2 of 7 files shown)


CUDACore/src/compiler/compilation.jl
```diff
@@ -288,11 +289,14 @@ end
...
-        ptxas_candidates = filter(ptx_sms) do sm
-            base_version(sm) >= base_version(minreq.sm) && runs_on(sm, capability(dev))
+        # Pick the most specific capability supported by the selected PTX ISA whose
+        # cubin would load on the current device.
+        supported = filter(sm -> base_version(sm) >= base_version(minreq.sm), ptx_sms)
+        ptxas_candidates = filter(sm -> runs_on(sm, capability(dev)), supported)
+        if isempty(ptxas_candidates) && _for_reflection
+            # Reflection stops at PTX, so it may use an older virtual target even when
+            # ptxas could not produce a cubin for this device from it.
+            ptxas_candidates = filter(sm -> base_version(sm) <= capability(dev), supported)
```
AntonOresten: This fallback lets `a`/`f` targets through: `base_version` strips the suffix, so on a CC 12.0 device with `ptx=v"8.0"` both sm_90 and sm_90a survive, and `argmax(sm_key)` prefers `:arch`. `code_ptx` then emits `.target sm_90a` where it used to emit `.target sm_90`. The fallback is justified by PTX forward compatibility, which only baseline targets have (compute_90a PTX is not supported on Blackwell). So maybe:

```julia
    ptxas_candidates = filter(supported) do sm
        sm.feature_set === :baseline && base_version(sm) <= capability(dev)
    end
```

CUDACore/src/compiler/sm.jl
```diff
@@ -104,19 +104,27 @@ cpu_name(sm::SMVersion) = "sm_$(sm.major)$(sm.minor)$(suffix(sm))"
...
-# `dev_cap`? Per NVIDIA's PTX ISA reference (.target directive):
-#   - baseline: forward-compatible (onion model) -- any sm_X runs on sm_Y for Y >= X.
-#   - family:   same architecture family (currently == same major) and forward-portable
-#               within the family.
-#   - arch:     locked to one exact CC; cubin only loads on devices with that exact cap.
+# Family-specific compatibility is not derivable from the compute-capability number.
+# Keep this table conservative: unlisted family targets are assumed to run only on their
+# exact capability until NVIDIA documents additional members.
+const family_compat = Dict(
+    v"10.0" => (v"10.3",),
```
AntonOresten: Nit: PTX ISA 9.3 Table 61 does define families by major version (sm_10x, sm_11x, sm_12x, each "plus future targets in the family"), and NVIDIA states the guarantee as same major with equal-or-higher minor. The only anomaly was sm_101f, since renamed sm_110f. Keeping a table is fine as a conservative choice, but the comment should say it's stricter than NVIDIA's rule, not that the rule isn't derivable. Note the branches now disagree on future minors: sm_103 would be accepted on the coming 10.7, while sm_103f would not.

What I mean is this would in the near future become something like:

```julia
v"10.0" => (v"10.3", v"10.7", v"10.9"),
v"10.3" => (v"10.7", v"10.9"),
v"10.7" => (v"10.9",),
```

Alternative that needs no upkeep and matches NVIDIA's rule:

```julia
sm.major == dev_cap.major && cap <= dev_cap && cap != v"10.1" && dev_cap != v"10.1"
```

CUDACore/src/compatibility.jl
```diff
@@ -10,8 +10,9 @@ const highest = v"999"
 #
 #   - Baseline (no suffix, e.g. sm_90): the forward-compatible feature set. Code compiled
 #     for sm_X runs on any sm_Y with Y >= X (onion model).
```
AntonOresten: Nit: "Code compiled for sm_X runs on any sm_Y with Y >= X" now reads as contradicting
`runs_on`, since that's only true of PTX.

maleadt:
Thanks, addressed comments.

PR #3272, opened 2026-09-14
Hand the GPUArrays-generic Enzyme rules to Enzyme.jl

vchuravy:
> Draft: blocked on Enzyme 0.13.204 being released. Depends on EnzymeAD/Enzyme.jl#3590.

Three groups of rules in `CUDACore/ext/EnzymeCoreExt.jl` have nothing to do with CUDA:

| Lines (before) | Definitions | The only CUDA-specific part |
|---|---|---|
| 634–792 | rules for `GPUArrays.mapreducedim!` and `GPUArrays._mapreduce` | the `Annotation{<:AnyCuArray{T}}` bound — the bodies are `fill!`, `.+=`, and a call back into the same GPUArrays function |
| 569–632 | `make_zero`/`make_zero!` + `float_only`/`bulk_zeroable`/`make_zero_via_host` | `DenseCuArray`, and `Array(prev)` for the host round-trip |
| 404–490 | the `Base.fill!` rules | the `MemsetCompatTypes` bound, which describes our memset implementation rather than anything the derivative needs |

The first group writes rules for another package's API outright. Keying all three on `DenseCuArray`/`AnyCuArray` meant no other backend got them — AMDGPU's `AMDGPUEnzymeCoreExt` has none of these, and `sum` over a `JLArray` aborted.

EnzymeAD/Enzyme.jl#3590 takes them over, keyed on `AbstractGPUArray`/`AnyGPUArray`, split across a new `EnzymeGPUArraysExt` (the reduction entry points need the full GPUArrays package) and the existing `EnzymeGPUArraysCoreExt`. This PR deletes the originals, along with the extension's now-unused `using GPUArrays`, and leaves a comment saying where they went.

**No coverage is lost.** `DenseCuArray` is `CuArray{T,N}` with no wrappers, so `AbstractGPUArray` is its exact analogue, and `AnyCuArray` maps onto `AnyGPUArray`. The `fill!` rules actually gain: their bound widens from `MemsetCompatTypes` to every float, complex and integer element type, so `Float64` and complex `CuArray`s stop falling through to the fill kernel.

**No transition hazard.** `AnyCuArray{T,N} <: AnyGPUArray{T,N}` holds including the wrapped-array arm, so as long as both sets exist these methods are strictly more specific and keep winning — no ambiguity either way round, and the two PRs can land in either order. What this PR needs is only that Enzyme ≥ 0.13.204 is *released* before it merges, otherwise the rules are simply absent for a while.

Four bugs in these rules were fixed on the way over rather than propagated to every backend; they're listed in EnzymeAD/Enzyme.jl#3590. Coverage on the Enzyme side is 26 new `JLArray` assertions, which run in CI with no device.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_018ofARZr5bm2nYMhzHTrHxK


CUDACore/ext/EnzymeCoreExt.jl
```diff
@@ -566,230 +477,13 @@ function EnzymeCore.EnzymeRules.noalias(::Type{CT}, ::UndefInitializer, args...)
...
-
-    return (nothing, nothing, nothing)
-end
+# `make_zero`/`make_zero!` for GPU arrays, the `Base.fill!` rules and the rules
+# for `GPUArrays.mapreducedim!`/`GPUArrays._mapreduce` used to live here, keyed on
+# `DenseCuArray`/`AnyCuArray`. Nothing in them was CUDA specific, so they moved to
+# Enzyme's GPUArraysCore and GPUArrays extensions, keyed on `AbstractGPUArray`,
+# where every backend gets them. The `fill!` rules covered only
+# `MemsetCompatTypes` here and now cover every float, complex and integer element
+# type. Requires Enzyme >= 0.13.204.
```
vchuravy, line 486:
```suggestion
```

Issue #2572, opened 2024-11-29
Multiplication with a symmetrically wrapped CuSparseMatrix returns wrong answer


mipals:
**Describe the bug**

Multiplication with a symmetrically wrapped CuSparseMatrix outputs the multiplication with the CuSparseMatrix itself and not the symmetric matrix. 

**To reproduce**

The Minimal Working Example (MWE) for this bug:

```julia
Id = [1, 2, 3, 4, 1, 4, 5, 2, 4, 6, 7, 3, 4, 8, 4, 5, 8, 9, 4, 5, 10]
Jd =  [1, 2, 3, 4, 5, 5, 5, 6, 6, 6, 7, 8, 8, 8, 9, 9, 9, 9, 10, 10, 10]
V = [1.0, 1.0, 1.0, 1.0, 2.0, 2.0, 1.0, 2.0, 2.0, 1.0, 1.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 1.0, 2.0, 2.0, 1.0]
T = sparse(Id,Jd,V)
v = ones(10)
Symmetric(T)*v - T*v # Non-zero
# However equivalent on the GPU does not work the same way
Tgpu = CuSparseMatrixCSR(T)
vgpu = CuVector(v)
Symmetric(Tgpu)*vgpu - Tgpu*vgpu # 0
```

<details><summary>Manifest.toml</summary>
<p>

```
Paste your Manifest.toml here, or accurately describe which version of CUDA.jl and its dependencies (GPUArrays.jl, GPUCompiler.jl, LLVM.jl) you are using.
```

</p>
</details>

**Expected behavior**

I expected the behaviour to be similar to the CPU version.

**Version info**

Details on Julia:

```
Julia Version 1.11.0
Commit 501a4f25c2b (2024-10-07 11:40 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 24 × Intel(R) Xeon(R) Gold 6126 CPU @ 2.60GHz
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, skylake-avx512)
Threads: 24 default, 0 interactive, 12 GC (on 24 virtual cores)
Environment:
  JULIA_EXCLUSIVE = 0
  LD_LIBRARY_PATH = /lsf/10.1/linux3.10-glibc2.17-x86_64/lib
[...]

PR #3279, opened 2026-09-16
Add diag and Diagonal for GPU sparse matrices

aalquwayfili:
As discussed in #2572

`diag` of a `CuSparseMatrix` fell back to the generic `AbstractArray` method, which indexes element by element and so failed with a scalar indexing error. `Diagonal` and `tr` are built on `diag`, so they failed too.

No new kernel is needed: in COO a diagonal is just the entries where `colInd - rowInd == k`, so this is a mask and a scatter. CSC and CSR convert to COO first, the same way `triu`, `tril` and `kron` in this file already do, and `Transpose`/`Adjoint` read the opposite diagonal of the parent.

One deliberate difference from the CPU: `diag` returns a dense `CuVector` where SparseArrays returns a `SparseVector`. That is so `tr` and `Diagonal` work, since reductions over `CuSparseVector` still fall back to scalar indexing. Happy to change it if you would rather match the CPU type.

This addresses the follow-up ask in #2572, not the original `Symmetric` question, which was answered as by design, so the issue should stay open.

Test: new testset in `lib/cusparse/test/linalg.jl` covering 3 formats, 4 element types, square and non-square shapes, offsets -4 to 4, transpose and adjoint, a matrix with nothing stored on the diagonal, and `Diagonal(A) * x`. Locally on a GPU: the whole cuSPARSE suite passes (10,568 assertions across 14 files), `linalg.jl` is 968/968 with the change, and 624 of those fail without it.


maleadt:
Thanks, this looks good. I pushed two small commits.
On the return type: agreed that a dense CuVector is the right choice here.

PR #3282, opened 2026-09-22
Add a fastmath option to CUDABackend

maleadt:
This adds a `fastmath` option to `CUDABackend`, so that KernelAbstractions kernels can opt into the same kernel-wide floating-point relaxations as `@cuda fastmath=true`: fast-math flags on every floating-point instruction, less precise square roots, and flushing of `Float32` subnormals to sign-preserving zero.

```julia
using CUDA, KernelAbstractions

@kernel function axpy!(y, a, x)
    i = @index(Global)
    @inbounds y[i] += a * x[i]
end

kernel! = axpy!(CUDABackend(fastmath=true), 256)
kernel!(y, 2f0, x; ndrange=length(y))
```

Like `@cuda`, the default follows Julia's `--math-mode` setting. The existing positional `CUDABackend(prefer_blocks, always_inline)` constructor keeps working, and the option is also forwarded to the compiler job used by Enzyme (`EnzymeCore.compiler_job_from_backend`).

An earlier version of this PR motivated the option with global-memory `Float32` atomic additions, which LLVM 23.1.1 lowered to a CAS loop in IEEE mode. That no longer applies: NVPTX_LLVM_Backend_jll 23.1.2 backports llvm/llvm-project#206154, so these atomics lower to the native (subnormal-flushing) `atom.add.f32` regardless of fast math, and the atomics tests and documentation have been dropped accordingly.


maleadt:
Oh wait, I just saw https://github.com/llvm/llvm-project/pull/206154 now. I'll just backport that to fix the performance regression, but this PR seems useful regardless to give people full control.

vchuravy (approved):

maleadt:
CI failure unrelated, will be handled by https://github.com/JuliaGPU/CUDA.jl/pull/3283.

